PR #6, opened 2021-06-24
add giant swarm deployment recording rules + change ingress alert

ubergesundheit:
Towards https://github.com/giantswarm/giantswarm/issues/17284

This PR attempts to solve:

- Current `IngressControllerDeploymentNotSatisfied` missing some deployments. Filtering on `deployment` breaks when customer chooses an own name or installs additional `nginx-ingress-controller-app` instances. We change to filtering on label `app: "nginx-ingress-controller"`.
- Notifying for non-giantswarm nginx ingress controller installations by introducing recording rules which select based on label `giantswarm.io/service-type: "managed"`

We tested this by 

- Changing the `App` CR on `ginger` to use the test catalog and `15df160` from this PR
- Installing the `nginx-ingress-controller-app` onto a WC on `ginger` with only a single node.
- The deployment requires two pod replicas on different nodes, thus the deployment was not satisfied.
- Alert worked

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/recording-rules/gs-managed-app-deployment-status.rules.yml
+++ b/helm/prometheus-rules/templates/recording-rules/gs-managed-app-deployment-status.rules.yml
@@ -9,11 +9,15 @@ spec:
   groups:
   - name: gs-managed-app-deployments
     rules:
-    - expr: kube_deployment_status_replicas_available *
-            on (deployment) group_left(label_giantswarm_io_service_type, label_app)
-            kube_deployment_labels{label_giantswarm_io_service_type="managed"}
+    - expr: label_replace(
+              kube_deployment_status_replicas_available *
+              on (deployment) group_left(label_app_kubernetes_io_name)
+              kube_deployment_labels{label_giantswarm_io_service_type="managed"},
+            "managed_app", "$1", "label_app_kubernetes_io_name", "(.*)" )
       record: managed_app_deployment_status_replicas_available
-    - expr: kube_deployment_status_replicas_unavailable *
-            on (deployment) group_left(label_giantswarm_io_service_type, label_app)
-            kube_deployment_labels{label_giantswarm_io_service_type="managed"}
+    - expr: label_replace(
+              kube_deployment_status_replicas_unavailable *
+              on (deployment) group_left(label_app_kubernetes_io_name)
+              kube_deployment_labels{label_giantswarm_io_service_type="managed"},
+            "managed_app", "$1", "label_app_kubernetes_io_name", "(.*)" )
       record: managed_app_deployment_status_replicas_unavailable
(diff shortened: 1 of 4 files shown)


ced0ps (commented):
LGTM
Waiting on Atlas review

helm/prometheus-rules/templates/recording-rules/gs-managed-app-deployment-status.rules.yml
```diff
@@ -0,0 +1,19 @@
...
+  name: gs-managed-app-deployment-status.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+  - name: gs-managed-app-deployments
+    rules:
+    - expr: kube_deployment_status_replicas_available *
+            on (deployment) group_left(label_giantswarm_io_service_type, label_app)
+            kube_deployment_labels{label_giantswarm_io_service_type="managed"}
+      record: kube_deployment_status_replicas_available_gs
```
QuentinBisson, line 15: Would it not make more sense to call this managed_app_deployment_status_replicas_available ?

helm/prometheus-rules/templates/recording-rules/gs-managed-app-deployment-status.rules.yml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+  name: gs-managed-app-deployment-status.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+  - name: gs-managed-app-deployments
+    rules:
+    - expr: kube_deployment_status_replicas_available *
+            on (deployment) group_left(label_giantswarm_io_service_type, label_app)
```
QuentinBisson, line 13: Can you not use the app label instead of label_app ? This should not be

ubergesundheit: `label_app` is a is a label (or field?) of the `kube_deployment_labels` metric which reflects the actual kubernetes label `app` of the deployment on the cluster.

Should we maybe rename the `label_app` to something else?

helm/prometheus-rules/templates/recording-rules/gs-managed-app-deployment-status.rules.yml
ubergesundheit: yeah good input! Thanks, will change!

helm/prometheus-rules/templates/recording-rules/gs-managed-app-deployment-status.rules.yml
QuentinBisson: If possible, I would because this is confusing with the app label on the metric. @giantswarm/team-atlas what do you think here ? Would it make sense to have a label called managed_app=app_name ?

ubergesundheit: yeah makes sense. Used `label_replace` to rename to `managed_app`

QuentinBisson (approved):

PR #17, opened 2021-07-07
Add paging alerts for docker and containerd being down

yulianedyalkova:
Towards https://github.com/giantswarm/adidas/issues/737.

Tested on `gorgoth` - [MC](https://giantswarm.app.opsgenie.com/alert/detail/253c85ba-d05a-4b57-8877-819f1d33d384-1625643077580/details) and [WC](https://giantswarm.app.opsgenie.com/alert/detail/1575e647-4964-4364-b1b5-74ef1017dace-1625652122970/details).

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/systemd.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/systemd.management-cluster.rules.yml
@@ -33,3 +33,14 @@ spec:
         severity: page
         team: biscuit
         topic: infrastructure
+    - alert: ManagementClusterRuntimeSystemdUnitFailed
+      annotations:
+        description: '{{`Runtime systemd unit {{ $labels.name }} is failed on {{ $labels.instance }}.`}}'
+        opsrecipe: critical-systemd-unit-failed/
+      expr: node_systemd_unit_state{cluster_type="management_cluster",name=~"docker.service|containerd.service", state="failed"} == 1
+      for: 5m
+      labels:
+        area: kaas
+        severity: page
+        team: biscuit
+        topic: infrastructure
--- a/helm/prometheus-rules/templates/alerting-rules/systemd.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/systemd.management-cluster.rules.yml
@@ -37,7 +37,7 @@ spec:
       annotations:
         description: '{{`Runtime systemd unit {{ $labels.name }} is failed on {{ $labels.instance }}.`}}'
         opsrecipe: critical-systemd-unit-failed/
-      expr: node_systemd_unit_state{cluster_type="management_cluster",name=~"docker.service|containerd.service", state="failed"} == 1
+      expr: node_systemd_unit_state{cluster_type="management_cluster",name=~"docker.service|containerd.service",state!="active"} == 1
       for: 5m
       labels:
         area: kaas
--- a/helm/prometheus-rules/templates/alerting-rules/systemd.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/systemd.management-cluster.rules.yml
@@ -38,7 +38,7 @@ spec:
         description: '{{`Runtime systemd unit {{ $labels.name }} is failed on {{ $labels.instance }}.`}}'
         opsrecipe: critical-systemd-unit-failed/
       expr: node_systemd_unit_state{cluster_type="management_cluster",name=~"docker.service|containerd.service",state!="active"} == 1
-      for: 5m
+      for: 15m
       labels:
         area: kaas
         severity: page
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/alerting-rules/systemd.management-cluster.rules.yml
```diff
@@ -33,3 +33,14 @@ spec:
         severity: page
         team: biscuit
         topic: infrastructure
+    - alert: ManagementClusterRuntimeSystemdUnitFailed
+      annotations:
+        description: '{{`Runtime systemd unit {{ $labels.name }} is failed on {{ $labels.instance }}.`}}'
+        opsrecipe: critical-systemd-unit-failed/
+      expr: node_systemd_unit_state{cluster_type="management_cluster",name=~"docker.service|containerd.service",state!="active"} == 1
+      for: 5m
```
MarcelMue, line 41: 5m seems a little low - docker could have PLEG issues which can take some time to resolve.

15m seems reasonable to me.

tfussell:
I think this works for detecting something like a misconfiguration or broken docker install, but I feel like the most common issue is docker not responding or OOM. Kubelet should report this condition via PLEG, but I just wanted to mention that we need to think about what kind of issues we want to detect with this.

yulianedyalkova:
> I think this works for detecting something like a misconfiguration or broken docker install, but I feel like the most common issue is docker not responding or OOM. Kubelet should report this condition via PLEG, but I just wanted to mention that we need to think about what kind of issues we want to detect with this.

Systemd does a lot more than that. Examples for docker (containerd is not that good but we can improve that when we move to pure containerd):
* `docker.service` has the following settings:
```
StartLimitBurst=3
StartLimitInterval=60s
```

which will bring the `docker` service into a failing state if it tries to start more than 3 times within a minute. I killed the dockerd process multiple times and ended up with a failing `docker` service that preserved for at least 11 mins (when I had to manually restart it):
```
● docker.service - Docker Application Container Engine
     Loaded: loaded (/run/systemd/system/docker.service; enabled; vendor preset: enabled)
    Drop-In: /etc/systemd/system/docker.service.d
             └─10-giantswarm-extra-args.conf
     Active: failed (Result: start-limit-hit) since Thu 2021-07-08 01:59:34 UTC; 11min ago
TriggeredBy: ● docker.socket

```

* `docker` service has a dependency of `docker.socket` and systemd internally performs healthchecks against the socket (killing `docker` process caused the socket unit to fail in the previous example). So if the socket disappears/goes unresponsive -> `docker` will go into failing state.

So my feeling is that this should be enough for a simple down check (or is at least better than the current state where we don't have anything).

kopiczko:
Is it possible that docker/containerd is not active and kubelet is not down? We are alerted by the kubelet being down already.

tfussell:
> Is it possible that docker/containerd is not active and kubelet is not down? We are alerted by the kubelet being down already.

We discussed this in our team a bit [here](https://gigantic.slack.com/archives/C01BYMF6RN0/p1625194028294500). I think it would be rare for kubelet to be up while Docker is down. That said, I could imagine a case where Docker is going OOM (fluentd often causes this), restarting, and kubelet works well enough that it doesn't alert. The restarts would still trigger this alert it sounds like. It also help locate issues more quickly.

We should definitely consider if we're going to end up with two pages when a node goes down as a result of this alert: `ServiceLevelBurnRateTooHigh` for kubelet and `WorkloadClusterRuntimeSystemdUnitFailed`.

kopiczko:
Thanks @tfussell for the detailed answer.

Do we have an efficient way to figure that Docker is OOM-ing because of a single container creating a pressure? It doesn't feel like this goes into symptom based alerting direction.

Also why not making this alert SLO-based?

yulianedyalkova:
I just tested again by killing `containerd` and it does bubble up to `ServiceLevelBurnRateTooHigh`.

> That said, I could imagine a case where Docker is going OOM (fluentd often causes this), restarting, and kubelet works well enough that it doesn't alert. The restarts would still trigger this alert it sounds like.

If it restarts more than 3 times within a minute and the service dies, that will get caught by both alerts. Otherwise it won't be caught by any. So more or less same behaviour.

>It also help locate issues more quickly.

That's true. I'm testing on `gorgoth` and the new alert came an hour ago where I still haven't received `ServiceLevelBurnRateTooHigh` cause it's just going up too slow (most likely because there is almost no load there right now).

It also makes debugging a little bit easier/faster cause it directly tells you what's broken instead of having to narrow it down to docker/containerd. 

I'm however totally fine with closing if it doesn't add that much value over `ServiceLevelBurnRateTooHigh`.

kopiczko:
> It also makes debugging a little bit easier/faster cause it directly tells you what's broken instead of having to narrow it down to docker/containerd.

I'm totally with you. But if we add more alerts that fire at the same time that may become overwhelming. Maybe extending the ops-recipe is an option?

I surely won't block this PR if you feel it's helpful.

tfussell:
Given the discussion, I am feeling like the risk of having multiple pages for the same issue outweighs the benefit of faster debugging. It feels reasonable to say that kubelet is responsible for monitoring docker/containerd. As a compromise, what if we make this a non-paging alert so we can find it in our alerts channel if needed?

PR #24, opened 2021-07-19
exclude already expired certificates from CertificateSecretWillExpireInLessThanTwoWeeks rule

ubergesundheit:
This PR:

Changes alert query of `CertificateSecretWillExpireInLessThanTwoWeeks` to exclude already expired certificates to reduce the amount of alerts.

Once, released, the `Only on Mondays: CertificateSecretWillExpireInLessThanTwoWeeks` global policy should be removed.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Exclude expired certificates in secrets for `CertificateSecretWillExpireInLessThanTwoWeeks`.
+
 ## [0.4.0] - 2021-07-14
 
 ### Changed
--- a/helm/prometheus-rules/templates/alerting-rules/certificate.all.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/certificate.all.rules.yml
@@ -14,7 +14,7 @@ spec:
       annotations:
         description: '{{`Certificate stored in Secret {{ $labels.namespace }}/{{ $labels.name }} on {{ $labels.cluster_id }} will expire in less than two weeks.`}}'
         opsrecipe: managed-app-cert-manager/
-      expr: (cert_exporter_secret_not_after{type="secret"} - time()) < 2 * 7 * 24 * 60 * 60
+      expr: (cert_exporter_secret_not_after{type="secret"} - time()) < 2 * 7 * 24 * 60 * 60 and cert_exporter_secret_not_after{type="secret"} > time()
       for: 5m
       labels:
         area: managedapps


calvix:
so there won't be any alert or already expired certs? is that something we want?

ubergesundheit:
> so there won't be any alert or already expired certs? is that something we want?

In theory no certificate should reach its expiration date if each alert is acted upon

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Exclude expired certificates in secrets for `CertificateSecretWillExpireInLessThanTwoWeeks`.
```
cokiengchiara, line 12:
```suggestion
- Exclude expired certificates in secrets for `CertificateSecretWillExpireInLessThanTwoWeeks`, to reduce operational toil of non-actionable, non GS-managed certificates.
```

cokiengchiara (approved):
Suggesting to add info on why we are doing it in changelog. Otherwise, LGTM if tested!

CHANGELOG.md
ubergesundheit: Thats not entirely true. This has nothing to do with non GS-managed certs

calvix:
> In theory no certificate should reach its expiration date if each alert is acted upon

then why we need to get rid of alerting on expired alerts? This feels weird, you say no alert should reach expiration yet you add condition to not alert on expired alerts

ubergesundheit:
> then why we need to get rid of alerting on expired alerts? This feels weird, you say no alert should reach expiration yet you add condition to not alert on expired alerts

Yes this sounds really wierd, I think in the past we failed to act upon some alerts, this is an atempt to recude alerts.

Expired certs are either unused (we can ignore these), or if in use, customer should have already told us

calvix:
in the end it's up to you, its alert for your team :) I was just curious 

not blocking

kopiczko:
Controversial. Do you have a particular example?

ubergesundheit:
On `orion / 2g2gc` in namespace `logging`, there is secret `elasticsearch-prod-certs`

```
$ k view-secret -n logging elasticsearch-prod-certs tls.crt | openssl x509 -noout -enddate
notAfter=Jul  7 08:35:46 2021 GM
```

also `kibana-prod-certs`

```
$ k view-secret -n logging kibana-prod-certs tls.crt | openssl x509 -noout -enddate
notAfter=Jul  7 08:35:55 2021 GMT
```

Both secrets have no `Certificate` CR and also seem otherwise to be unused.

Maybe these are the only examples and the whole idea is not as good, I certainly see the disconnect between excluding expired certs and also expecting people to act in the two weeks the alert fires between actually expiring.

JosephSalisbury:
I think this is more a symptom of alerting for certificates we don't care about, right?

kopiczko:
Don't those certs belong to the customer?

ubergesundheit:
> Don't those certs belong to the customer?

As of now this is a really hard problem (giantswarm/giantswarm#16879). We could change cert-exporter only to alert on certificates issued by let's encrypt, but that would exclude self-signed certificates (we use for example for kiam). But then we would also get notified for certificates issued by non-gs cert-manager installations.

Once https://github.com/jetstack/cert-manager/pull/3828 is released, we would be able to filter for certificate secrets which are specially labeled with a managed-by-giantswarm label and only alert for these, but only for clusters with the updated platform releases (on AWS), this will also require some changes in cert-exporter (which is also part of releases)

whites11:
My 2cc is that expired certificates should be a rare exception to the rule.
We should use the `silence` repo to handle those rarities.
Silencing the alert like this smells like a very, very bad idea to me.

kopiczko:
@ubergesundheit thanks, I understand the problem now. I won't block it if that helps. Possibly @whites11 suggestion could help.

ced0ps (approved):
LGTM

PR #28, opened 2021-07-21
Add KiamSTSIssuingErrors

njuettner:
This hopefully will block `ExternalDNSCantAccessRegistry` paging during the night, by adding STSIssue alert.

`ExternalDNSCantAccessRegistry` uses `cancel_if_kiam_has_errors`, see
 https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/alerting-rules/external-dns.rules.yml#L26, which should be inhibited here https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/alerting-rules/inhibit.all.rules.yml#L32-L40.
### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/kiam.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/kiam.rules.yml
@@ -33,7 +33,7 @@ spec:
       annotations:
         description: '{{`Kiam pod {{ $labels.namespace}}/{{ $labels.pod_name }} on {{ $labels.cluster_id}}/{{ $labels.cluster }} has increased STS issuing errors.`}}'
         opsrecipe: kiam_sts_issuing_errors_total/
-      expr: increase(kiam_sts_issuing_errors_total[10m]) > 0
+      expr: increase(kiam_sts_issuing_errors_total[10m]) > 5
       for: 15m
       labels:
         area: managedservices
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/kiam.rules.yml
```diff
@@ -27,4 +27,21 @@ spec:
         severity: page
         team: firecracker
         topic: kiam
+  - name: kiam
+    rules:
+    - alert: KiamSTSIssuingErrors
+      annotations:
+        description: '{{`Kiam pod {{ $labels.namespace}}/{{ $labels.pod_name }} on {{ $labels.cluster_id}}/{{ $labels.cluster }} has increased STS issuing errors.`}}'
+        opsrecipe: kiam_sts_issuing_errors_total/
+      expr: increase(kiam_sts_issuing_errors_total[10m]) > 0
```
whites11, line 36: Maybe we could allow some errors? Like 5 or 10 dunno

whites11 (approved):
Otherwise LGTM

helm/prometheus-rules/templates/alerting-rules/kiam.rules.yml
njuettner, line 36:
```suggestion
      expr: increase(kiam_sts_issuing_errors_total[10m]) > 5
```

njuettner: Makes sense 👍🏻

PR #30, opened 2021-07-21
alert if felix cannot save or restore iptables rules

glitchcrab:
towards https://github.com/giantswarm/giantswarm/issues/18140

This PR:

- Adds alerts if Calico is failing to update or restore iptables rules in KVM clusters.

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/calico.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/calico.rules.yml
@@ -35,3 +35,35 @@ spec:
         severity: notify
         team: ludacris
         topic: kubernetes
+{{- if eq .Values.managementCluster.provider.kind "kvm" }}
+    - alert: CalicoNodeFailingToSaveIptables
+      annotations:
+        description: '{{`calico-node {{ $labels.namespace}}/{{ $labels.pod_name }} on {{ $labels.cluster_id}}/{{ $labels.cluster }} is failing to save iptables rules.`}}'
+        opsrecipe: calico-iptables-failing/
+      expr: increase(felix_iptables_save_errors[10m]) > 0
+      for: 10m
+      labels:
+        area: kaas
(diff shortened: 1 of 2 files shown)


glitchcrab:
I'm unsure on a couple of things here:

- should this be for KVM only? it's the only place we use calico as the CNI.
- should it stay at notify severity for now? I don't know how much noise it will create.
- do we think the time span to trigger is sensible?

helm/prometheus-rules/templates/alerting-rules/calico.rules.yml
```diff
@@ -35,3 +35,35 @@ spec:
...
+        opsrecipe: calico-iptables-failing/
+      expr: increase(felix_iptables_save_errors[10m]) > 0
+      for: 10m
+      labels:
+        area: kaas
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "true"
+        severity: notify
```
MarcelMue, line 51: Just notify for now?

MarcelMue (approved):

helm/prometheus-rules/templates/alerting-rules/calico.rules.yml
glitchcrab: yeah, i'm unsure as to how much noise it will create - my plan was to see how it goes and then bump the severity once we're happy

tfussell:
1. We don't run felix on other providers, but it wouldn't hurt to restrict it to KVM only.
2. I'm not sure if these errors occur sometimes during normal functioning. Keeping it as notify will let us see how often it happens and decide if it should be a paging alert.
3. 10 minutes seems fine. We can adjust it if we find that only multiple errors in a short time are indicative of a serious problem.

tfussell (approved):

PR #32, opened 2021-07-22
Add recording in recording Prometheus Rules to avoid conflicts with alerting rules

QuentinBisson:
This PR:

- Adds recording in recording Prometheus Rules to avoid conflicts with alerting rules.

We need this because the service level recording rules overwrite the service level alerting rules.

giphy

### Checklist


JosephSalisbury (approved):
lgtmiiwfy

PR #41, opened 2021-07-30
Add basic Falco rules

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/13754

This PR adds prometheus alerts for Falco events

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/falco.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/falco.rules.yml
@@ -60,3 +60,15 @@ spec:
         severity: notify
         team: ludacris
         topic: security
+    - alert: FalcoAlwaysAlert
+      annotations:
+        description: '{{`Always true. How do I test prometheus-rules?`}}'
+      expr: (abs(0) = 0)
+      labels:
+        area: kaas
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        severity: notify
+        team: ludacris
+        topic: security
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/falco.rules.yml
```diff
@@ -0,0 +1,62 @@
...
+  name: falco.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+  - name: falco
+    rules:
+    - alert: FalcoCriticalAlertFiring
+      annotations:
+        description: '{{`Pod {{ $labels.k8s_ns_name }}/{{ $labels.k8s_pod_name }} triggered the Falco rule {{ $labels.rule }}.`}}'
+      expr: increase(falco_events{priority=~"0|1|2|3"}[10m] ) > 0
```
stone-z, line 17: Open to suggestions from those more knowledgeable. Falco events are atomic, so there were tradeoffs getting to this expression.
Originally, the idea was to page on any non-zero event count (with the optimistic goal that we get to 0 false positives on MCs). The downside there is that the alert fires forever once it fires once and "new" events occurring later are not meaningfully distinguished from the "old" ones.
Using `increase()` has the downside that it could miss events outside the sliding window (e.g. if prom is down) and that alerts will self-close after some time no matter what.

Is there a better way? Ultimately, we probably won't use the existing prom + alertmanager setup for Falco alerting in the long term, but we'll do this for now.

stone-z: Maybe @giantswarm/team-atlas has a better idea?

MarcelMue (approved):
Works for me

stone-z:
Going with this for now -- please holler if this can be improved

PR #44, opened 2021-08-04
turn firecracker daemonset alert into an SLO alert

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/17885

This PR:

- Turns `DaemonSetNotSatisfiedFirecracker` into `ServiceLevelBurnRateTooHigh` SLO alert.

We have a lot more instances of `DaemonSetNotSatisfied` alerts. However, let's try with this one and see if it works better as an SLO alert. Right now the availability is 99% and we only count daemonsets that have been not satisfied for 10 minutes as error. Furthermore, we do not count on new or unschedulable nodes. The original alert pages after 30 minutes so there is a chance that this one is actually more sensitive than before. We may have to adjust it.

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
+++ b/helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
@@ -38,6 +38,45 @@ spec:
         service: api-server
       record: slo_target
 
+      # -- daemonset
+    - expr: |
+        label_replace(
+          kube_daemonset_status_desired_number_scheduled{cluster_id!="argali|giraffe", namespace=~"giantswarm|kube-system", daemonset=~"aws-node|kiam-server|kiam-agent"},
+        "service", "$1", "workload_name", "(.*)" )
+      labels:
+        class: MEDIUM
+        area: kaas
+      record: raw_slo_requests
+      # -- the errors are counted as follows: 
+      # -- pods in a daemonset that are UNAVAILABLE NOW and have been UNAVAILABLE 10 MINUTES AGO 
+      # -- which are on a SCHEDULABLE node that was CREATED AT LEAST 10 MINUTES AGO
+    - expr: |
+        (
+          (
+            label_replace(
+              kube_daemonset_status_number_unavailable{cluster_id!="argali|giraffe", namespace=~"giantswarm|kube-system", daemonset=~"aws-node|kiam-server|kiam-agent"},
+              "service", "$1", "workload_name", "(.*)" ) > 0
+            and on (daemonset, node)
+            label_replace(
+              kube_daemonset_status_number_unavailable{cluster_id!="argali|giraffe", namespace=~"giantswarm|kube-system", daemonset=~"aws-node|kiam-server|kiam-agent"} offset 10m,
+              "service", "$1", "workload_name", "(.*)" ) > 0
+          )
+          and
+          on (node) kube_node_spec_unschedulable == 0
+        )
+        and
+        on (node) time() - kube_node_created > 10 * 60
+      labels:
+        class: MEDIUM
+        area: kaas
+      record: raw_slo_errors
+      # -- 99% availability
+      # -- this expression collects all the daemonsets and assigns the same slo target to all of them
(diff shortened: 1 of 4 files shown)


helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
```diff
@@ -38,6 +38,45 @@ spec:
...
+      # -- which are on a SCHEDULABLE node that was CREATED AT LEAST 10 MINUTES AGO
+    - expr: |
+        (
+          (
+            label_replace(
+              kube_daemonset_status_number_unavailable{cluster_id!="argali|giraffe", namespace=~"giantswarm|kube-system", daemonset=~"aws-node|kiam-server|kiam-agent"},
+              "service", "$1", "workload_name", "(.*)" ) > 0
+            and on (daemonset, node)
+            label_replace(
+              kube_daemonset_status_number_unavailable{cluster_id!="argali|giraffe", namespace=~"giantswarm|kube-system", daemonset=~"aws-node|kiam-server|kiam-agent"} offset 10m,
```
calvix, line 61: if we still ignore china we should add `akita`

helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
```diff
@@ -38,6 +38,45 @@ spec:
...
+          kube_daemonset_status_desired_number_scheduled{cluster_id!="argali|giraffe", namespace=~"giantswarm|kube-system", daemonset=~"aws-node|kiam-server|kiam-agent"},
+        "service", "$1", "workload_name", "(.*)" )
+      labels:
+        class: MEDIUM
+        area: kaas
+      record: raw_slo_requests
+      # -- the errors are counted as follows: 
+      # -- pods in a daemonset that are UNAVAILABLE NOW and have been UNAVAILABLE 10 MINUTES AGO 
+      # -- which are on a SCHEDULABLE node that was CREATED AT LEAST 10 MINUTES AGO
+    - expr: |
```
anvddriesch, line 53: The expression is a bit messy. I hope the comment helps.

helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
calvix: to all expressions where is this used

calvix (approved):
lgtm i assume its tested somehow :)

helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
anvddriesch: Oh yeah, you are right.
Although, does that even still make sense? The distinction only works for management-clusters anyway so I'm not sure if it's helpful at all.

calvix: yeah maybe its useless anyway :p

helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
```diff
@@ -38,6 +38,45 @@ spec:
         service: api-server
       record: slo_target
 
+      # -- daemonset
+    - expr: |
+        label_replace(
+          kube_daemonset_status_desired_number_scheduled{cluster_id!="argali|giraffe", namespace=~"giantswarm|kube-system", daemonset=~"aws-node|kiam-server|kiam-agent"},
```
njuettner, line 44: Do we also want to exclude akita?

anvddriesch: Right now it is not excluded so I just kept it that way. I personally think that we don't really need that distinction anymore anyway, especially because it does not affect workload clusters. Let's see if we see issues with `akita` that makes us think differently.

anvddriesch:
I tested it on `gauss` but can not say for sure that this won't be flaky at first. For now I will merge it and keep an eye on it for the rest of the day to make sure people won't get annoyed in the night.

PR #59, opened 2021-08-11
Add alert for blackbox exporter endpoint down

tfussell:
Towards https://github.com/giantswarm/telekom-panamax/issues/104

This PR:

- Adds an alerting rule for blackbox exporter endpoint down on KVM.

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
@@ -12,6 +12,17 @@ spec:
   groups:
   - name: kvm
     rules:
+    - alert: BlackboxExporterEndpointDown
+      annotations:
+        description: '{{`Blackbox exporter endpoint ({{ $labels.instance }}) is down.`}}'
+        opsrecipe: blackbox-exporter-endpoint-down/
+      expr: up{job="management-cluster-blackbox-ingress-http"} == 0
+      for: 15m
+      labels:
+        area: kaas
+        severity: page
+        team: rocket
+        topic: observability
     - alert: ManagementClusterPodPendingFor15Min
       annotations:
         description: '{{`Pod {{ $labels.namespace }}/{{ $labels.pod }} is stuck in Pending.`}}'
--- a/helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
@@ -23,6 +23,17 @@ spec:
         severity: page
         team: rocket
         topic: observability
+    - alert: BlackboxExporterEndpointNotResponding
+      annotations:
+        description: '{{`Blackbox exporter endpoint ({{ $labels.instance }}) is not responding.`}}'
+        opsrecipe: blackbox-exporter-endpoint-down/
+      expr: sum by(instance) (avg_over_time(probe_http_duration_seconds[5m])) > 0.2
+      for: 5m
+      labels:
+        area: kaas
+        severity: notify
+        team: rocket
+        topic: observability
     - alert: ManagementClusterPodPendingFor15Min
       annotations:
         description: '{{`Pod {{ $labels.namespace }}/{{ $labels.pod }} is stuck in Pending.`}}'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add recording rule for `grafana_analytics_sessions_total`
+
 ### Changed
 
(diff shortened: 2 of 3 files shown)


helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
```diff
@@ -12,6 +12,17 @@ spec:
   groups:
   - name: kvm
     rules:
+    - alert: BlackboxExporterEndpointDown
+      annotations:
+        description: '{{`Blackbox exporter endpoint ({{ $labels.instance }}) is down.`}}'
+        opsrecipe: blackbox-exporter-endpoint-down/
```
tfussell, line 18: I'll add this Ops Recipe while I wait for review.

tfussell: Recipe here https://github.com/giantswarm/prometheus-rules/pull/59

MarcelMue (approved):

helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
MarcelMue: The link to the recipe is this PR and not the recipe PR :D

wind0r:
15 minutes of one probe being down will "never" report any of the issues we currently have with kvm.

A Screenshot of the outage/issue from today shows quite good that this rule will not help to spot those problems
image

I think something like `if $x request in 5 minutes are over $y ms` would be better.
Avg blackbox response time is 4ms

wind0r:
@tfussell Just, I would have expected Giant Swarm to handle this. Some nginx pods didn't accept any traffic. Some info here: https://github.com/giantswarm/telekom-panamax/issues/106

This kinda was the first or second time nginx issues where spotted via Blackbox Exporter. Most of the time we spot broken overlay network.

The good part with those blackbox exporters is that "just" Giant Swarm stuff is involved. It can't be an application level issue on our side, since they just check if nginx is accessible.
If Loadbalancers cant access the cluster or responses are slow in my opinion it is something Giant Swarm should handle.

helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
```diff
@@ -12,6 +12,28 @@ spec:
...
+      expr: up{job="management-cluster-blackbox-ingress-http"} == 0
+      for: 15m
+      labels:
+        area: kaas
+        severity: page
+        team: rocket
+        topic: observability
+    - alert: BlackboxExporterEndpointNotResponding
```
tfussell, line 26: Added a new alert that actually uses the metrics reported by the blackbox exporter.

tfussell:
> @tfussell Just, I would have expected Giant Swarm to handle this. Some nginx pods didn't accept any traffic. Some info here: [giantswarm/telekom-panamax#106](https://github.com/giantswarm/telekom-panamax/issues/106)

I see. Thank you for the additional info. I added a new alert which should cover this use case. It is not configured to page us currently as we need to determine the correct alerting thresholds. We will check how often it is firing next week and make it paging once we have a good sense of what the threshold should be.

helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
```diff
@@ -12,6 +12,28 @@ spec:
...
+      annotations:
+        description: '{{`Blackbox exporter endpoint ({{ $labels.instance }}) is not responding.`}}'
+        opsrecipe: blackbox-exporter-endpoint-down/
+      expr: sum by(instance) (avg_over_time(probe_http_duration_seconds[5m])) > 0.2
```
pipo02mix, line 30: Out of curiosity, why 0.2 secs?

pipo02mix (approved):
it seems fair

helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml
tfussell: It was chosen arbitrarily based on the last few days of metrics. I wanted it to be high enough that normal variation didn't cause it to fire, but real issues would. I figure we can adjust it downwards if we miss a problem we should have been paged for or upwards if we're getting spurious alerts.

tfussell:
I will merge tomorrow morning just in case there are spurious pages.

webwurst:
@wind0r Can you point us to more incidents to test our query? For now I have `2021-08-13 10:00 - 16:00` for example.

wind0r:
@webwurst the other 2 nginx incidents are over 30 days old so data is gone :( sorry

ubergesundheit:
@tfussell any progress regarding the alert query?

tfussell:
> @tfussell any progress regarding the alert query?

I now have a rule which will reliably indicate the August 13th issue based on the historical data. I have asked for help from SIG Monitoring making sure it is performant. Tomorrow I will try to reproduce the flannel downtime to ensure it would catch that problem as well.

tfussell:
Closing in favor of https://github.com/giantswarm/prometheus-rules/pull/94

PR #73, opened 2021-08-19
drop user labels from `aggregation:grafana_analytics_sessions_total`

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/18571

This PR:

- adds/changes/removes etc

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/recording-rules/cortex.rules.yml
+++ b/helm/prometheus-rules/templates/recording-rules/cortex.rules.yml
@@ -212,6 +212,13 @@ spec:
       record: aggregation:dex_requests_status_ok
     - expr: grafana_analytics_sessions_total
       record: aggregation:grafana_analytics_sessions_total
+      labels:
+        user_locale: "none"
+        user_login: "none"
+        user_name: "none"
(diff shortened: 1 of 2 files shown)


QuentinBisson (approved):

helm/prometheus-rules/templates/recording-rules/cortex.rules.yml
```diff
@@ -212,6 +212,13 @@ spec:
       record: aggregation:dex_requests_status_ok
     - expr: grafana_analytics_sessions_total
       record: aggregation:grafana_analytics_sessions_total
+      labels:
+        user_locale: "none"
```
QuentinBisson, line 216: Should we not replace it with "" or null ?

TheoBrigitte: Doc says "A label with an empty label value is considered equivalent to a label that does not exist."
source: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

TheoBrigitte:
Tested successfully on ghost 

image

PR #94, opened 2021-09-03
Add blackbox exporter SLO recording rules

tfussell:
Towards https://github.com/giantswarm/telekom-panamax/issues/104

This PR:

- Adds SLO alerts for endpoints exposed by the blackbox-exporter (if it exists)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add recording rule for `grafana_analytics_sessions_total`
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


tfussell:
I tested this alert on gorgoth by deploying a temporary blackbox-exporter in the MC, setting up the worker service as a scrape target using `config`, and manipulating the WC to simulate various past incidents. I tested the following:
1. Change nginx ingress replicas to 0 in the WC, then back to 1
2. Break flannel by changing to a bad image and restarting a node
3. Break a node by disabling kubelet

Here is a graph of the results:

Screen Shot 2021-09-03 at 11 40 49 AM

tfussell:
@webwurst @ubergesundheit Do you think this covers the nginx ingress incident adequately?

webwurst:
Very cool! Thank you very much :) I can't judge how precisely this matches of the issues we saw in the past. But when we have the rule and alerting in place we can adjust threshold and values over time.

webwurst (approved):

PR #104, opened 2021-09-16
Remove labels from group_left for SLO alert

tfussell:
This PR:

- attempts to resolve the problem reported [in Slack](https://gigantic.slack.com/archives/C0E0V16DC/p1631628887005100) when blackbox-exporter SLO alert fires for multiple clusters at once

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
@@ -16,7 +16,6 @@ spec:
         opsrecipe: service-level-burn-rate-too-high/
       expr: |
         (
-        (
           slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
         and
           slo_errors_per_request:ratio_rate5m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
@@ -26,7 +25,6 @@ spec:
           slo_errors_per_request:ratio_rate6h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
         and
           slo_errors_per_request:ratio_rate30m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
-        )
         )
       for: 5m
       labels:
--- a/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
@@ -15,16 +15,18 @@ spec:
         description: '{{`Service level burn rate is too high for {{ $labels.service }} service.`}}'
         opsrecipe: service-level-burn-rate-too-high/
       expr: |
-        slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"}
-        and on (service)
         (
-        slo_errors_per_request:ratio_rate1h > on (service) group_left slo_threshold_high
+        (
+          slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
         and
-        slo_errors_per_request:ratio_rate5m > on (service) group_left slo_threshold_high
+          slo_errors_per_request:ratio_rate5m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
+        )
         or
-        slo_errors_per_request:ratio_rate6h > on (service) group_left slo_threshold_low
+        (
+          slo_errors_per_request:ratio_rate6h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
         and
-        slo_errors_per_request:ratio_rate30m > on (service) group_left slo_threshold_low
+          slo_errors_per_request:ratio_rate30m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
+        )
         )
       for: 5m
       labels:
--- a/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
@@ -18,13 +18,13 @@ spec:
         slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"}
         and on (service)
         (
-        slo_errors_per_request:ratio_rate1h > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_high
+        slo_errors_per_request:ratio_rate1h > on (service) group_left slo_threshold_high
         and
-        slo_errors_per_request:ratio_rate5m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_high
+        slo_errors_per_request:ratio_rate5m > on (service) group_left slo_threshold_high
         or
-        slo_errors_per_request:ratio_rate6h > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+        slo_errors_per_request:ratio_rate6h > on (service) group_left slo_threshold_low
         and
-        slo_errors_per_request:ratio_rate30m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+        slo_errors_per_request:ratio_rate30m > on (service) group_left slo_threshold_low
         )
       for: 5m
       labels:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Adjust `ServiceLevelBurnRateTooHigh` metric to avoid "multiple matches for labels" error.
+
 ## [0.23.0] - 2021-09-16
 
 ### Changed


helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -18,13 +18,13 @@ spec:
         slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"}
         and on (service)
         (
-        slo_errors_per_request:ratio_rate1h > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_high
```
tfussell, line 21: @giantswarm/sig-monitoring What do you think about this solution? It gets rid of the error when I run it in the browser, but I don't know if it affects something else inadvertently. `slo_threshold_high` doesn't have any labels.

whites11: I guess only @giantswarm/team-atlas can answer this question. I have no clue

QuentinBisson: I'm actually not sure this changes anything as the rules are defined in a tenant cluster prometheus anyway. The class should stay though

tfussell: > I'm actually not sure this changes anything as the rules are defined in a tenant cluster prometheus anyway.

@QuentinBisson Could you elaborate on this?

QuentinBisson:
Yes of course and sorry I cannot help. I will check with @giantswarm/team-atlas who can help here.
So basically, the alerting rule should already be triggered in a workload cluster prometheus or a management cluster prometheus so the prometheus instance knows which cluster it's monitoring so the cluster_id and type might not really be needed in the alert.

Regarging the issue, did you add some SLO alerts or components recently that could be missing those labels ?

JosephSalisbury:
This turned out to be an issue with Helm, right?

TheoBrigitte:
Could we please summarize this issue? What is the actual issue here?

On our side we found out that the `ServiceLevelBurnRateTooHigh` alert was actually missing in many installations since 26th of July and got fixed on 27th of September, it was annouced in [#news-incidents](https://gigantic.slack.com/archives/CDJJ17U81/p1631891777007900) and we track it here: https://github.com/giantswarm/giantswarm/issues/18975.

tfussell:
> Could we please summarize this issue? What is the actual issue here?

I believe there were two separate issues. The first was that the alert CR was actually missing in many places as you said and we're starting to be alerted now that it's fixed so that's good.

The second issue is that prometheus returns an error when running the SLO query in certain cases and I believe this may prevent other `ServiceLevelBurnRateTooHigh` alerts from firing. This error only occurs when the SLO threshold is exceeded so it's difficult to test. We had one service with a 100% error rate. Running https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml#L18-L28 in Prometheus would always return an error so I don't think the alert would ever fire.

As for the fix, if we look at `slo_target` (https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/recording-rules/service-level.rules.yml#L150-L154), it is a vector with only `area` and `service` labels. We are trying to `group_left` on `cluster_id` which doesn't exist on `slo_target` so we get `Error executing query: multiple matches for labels: grouping labels must ensure unique matches` sometimes. I will send you a URL where this happens so you can see it.

TheoBrigitte:
Managed to nail down the issue using following time frame: `2021-09-28 02:50:09` and `1d` range.

The issue arise when a service (`blackbox-exporter` here) fails on different cluster (`aaaa`, `bbbb`, and `cccc`)
image
So trying to group the above with the following ...
image
... result in this error
image

This fails because we are strictly asking to group on `cluster_type`, `cluster_id`, and `class`, which Prometheus can't do since we have 3 different `cluster_id`. Leaving the `group_left` empty allow Prometheus to group what is possible and fixes the issue.

whites11 (approved):
LGTM when tested properly

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -18,13 +18,13 @@ spec:
...
+        slo_errors_per_request:ratio_rate1h > on (service) group_left slo_threshold_high
```
TheoBrigitte, line 21:
```suggestion
        (
          slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
```

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -18,13 +18,13 @@ spec:
...
         and
-        slo_errors_per_request:ratio_rate5m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_high
+        slo_errors_per_request:ratio_rate5m > on (service) group_left slo_threshold_high
```
TheoBrigitte, line 23:
```suggestion
          slo_errors_per_request:ratio_rate5m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
        )
```

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -18,13 +18,13 @@ spec:
...
         or
-        slo_errors_per_request:ratio_rate6h > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+        slo_errors_per_request:ratio_rate6h > on (service) group_left slo_threshold_low
```
TheoBrigitte, line 25:
```suggestion
        (
          slo_errors_per_request:ratio_rate6h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
```

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -18,13 +18,13 @@ spec:
...
         and
-        slo_errors_per_request:ratio_rate30m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+        slo_errors_per_request:ratio_rate30m > on (service) group_left slo_threshold_low
```
TheoBrigitte, line 27:
```suggestion
          slo_errors_per_request:ratio_rate30m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
        )
```

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -18,13 +18,13 @@ spec:
         slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"}
         and on (service)
```
TheoBrigitte, line 19:
```suggestion
```

TheoBrigitte (changes requested):
Removing the `group_left` label list is correct, but I would also like to fix this expression and make service filtering right.

I was testing on `puma` and Prometheus restarted and trashed all the data, so you only have my words now that this works.

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -15,16 +15,18 @@ spec:
         description: '{{`Service level burn rate is too high for {{ $labels.service }} service.`}}'
         opsrecipe: service-level-burn-rate-too-high/
       expr: |
-        slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"}
-        and on (service)
         (
```
TheoBrigitte, line 18:
```suggestion
```

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -15,16 +15,18 @@ spec:
...
+        )
         or
-        slo_errors_per_request:ratio_rate6h > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+        (
+          slo_errors_per_request:ratio_rate6h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
         and
-        slo_errors_per_request:ratio_rate30m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+          slo_errors_per_request:ratio_rate30m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
+        )
         )
```
TheoBrigitte, line 30:
```suggestion
```

TheoBrigitte (approved):
Perfect :ok_hand: thanks you sir.

Those parenthesis can also be remove then.

tfussell:
Pinging basically everyone at GS since this affects everything that uses SLO alerting. Sorry for the noise.

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -15,16 +15,16 @@ spec:
...
+          slo_errors_per_request:ratio_rate5m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
+        )
         or
-        slo_errors_per_request:ratio_rate6h > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+        (
+          slo_errors_per_request:ratio_rate6h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
         and
-        slo_errors_per_request:ratio_rate30m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+          slo_errors_per_request:ratio_rate30m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
         )
```
tfussell, line 28: Seems like we could filter `{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"}` on the outer parentheses rather than on each metric. This would probably be less efficient though because data would be thrown away.

TheoBrigitte: Prometheus does not like this

image

TheoBrigitte:
Tested on `puma` :heavy_check_mark: 

Old alert yielding false positive cluster `bbbb` and `cccc`
image

New alert correctly firing only for cluster `aaaa`
image

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
tfussell: Oh well :)

helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
```diff
@@ -15,16 +15,16 @@ spec:
...
-        slo_errors_per_request:ratio_rate5m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_high
+          slo_errors_per_request:ratio_rate5m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_high
+        )
         or
-        slo_errors_per_request:ratio_rate6h > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+        (
+          slo_errors_per_request:ratio_rate6h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
         and
-        slo_errors_per_request:ratio_rate30m > on (service) group_left (cluster_type, cluster_id, class) slo_threshold_low
+          slo_errors_per_request:ratio_rate30m{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_left slo_threshold_low
```
MarcelMue, line 27: I am clearly missing some context because this list of services seems a bit arbitrary to me?

TheoBrigitte: I do not know where those services comes from, and they should probably excluded from the `raw_slo_errors`/`raw_slo_requests` recording rules rather than here, but we just moved them around.

anvddriesch: I think the correct way to exclude managed apps from the SLO stuff should be to remove the label `giantswarm.io/monitoring_basic_sli: "true"`. At least that's what I get from the documentation.

tfussell: I added a note in the Team Atlas board about improving this. Hope that's sufficient.

anvddriesch (approved):
Works for me!

PR #157, opened 2021-11-15
Reduce blackbox exporter SLO from 99.99% to 99.9%

kopiczko:
Fixes https://github.com/giantswarm/giantswarm/issues/19667


kopiczko:
See https://github.com/giantswarm/giantswarm/issues/19667#issuecomment-968928044

PR #161, opened 2021-11-23
Route alerts based on the provider

fiunchinho:
We have alerts like `RootVolumeSpaceTooLow` that are being routed to team phoenix, but they could be coming from a KVM installation. With these changes we route some alerts based on the provider.

This PR:

- Route alerts based on the provider

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/disk.workload-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/disk.workload-cluster.rules.yml
@@ -21,7 +21,11 @@ spec:
         area: kaas
         cancel_if_outside_working_hours: "true"
         severity: page
+        {{- if or (eq .Values.managementCluster.provider.kind "aws") (eq .Values.managementCluster.provider.kind "azure") }}
         team: phoenix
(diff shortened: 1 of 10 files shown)


whites11 (approved):

helm/prometheus-rules/templates/alerting-rules/disk.workload-cluster.rules.yml
```diff
@@ -21,7 +21,13 @@ spec:
         area: kaas
         cancel_if_outside_working_hours: "true"
         severity: page
+        {{- if eq .Values.managementCluster.provider.kind "aws" }}
```
tfussell, line 24: Why not use the same syntax as for rocket? i.e. `{{- else if or (eq .Values.managementCluster.provider.kind "aws") (eq .Values.managementCluster.provider.kind "azure") }}`

fiunchinho: yeah let's do that. Take a look now.

njuettner (approved):
lgtm

tfussell (approved):

PR #165, opened 2021-11-29
Route more KaaS alerts based on the provider

fiunchinho:
This PR:

- Route more KaaS alerts based on the provider
- Renamed route to use "KaaS" instead of "Phoenix"

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/deployment.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/deployment.management-cluster.rules.yml
@@ -64,7 +64,7 @@ spec:
         severity: page
         team: honeybadger
         topic: managementcluster
-    - alert: DeploymentNotSatisfiedPhoenix
+    - alert: DeploymentNotSatisfiedKaas
       annotations:
         description: '{{`Deployment {{ $labels.namespace}}/{{ $labels.deployment }} is not satisfied.`}}'
         opsrecipe: deployment-not-satisfied/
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/alerting-rules/deployment.management-cluster.rules.yml
```diff
@@ -64,7 +64,7 @@ spec:
         severity: page
         team: honeybadger
         topic: managementcluster
-    - alert: DeploymentNotSatisfiedPhoenix
+    - alert: DeploymentNotSatisfiedKaas
```
fiunchinho, line 67: I used `KaaS` instead of the team name.

TheoBrigitte (commented):
In case the provider is missing for some reason, this alert will silently stop working. That's bad.

fiunchinho:
> In case the provider is missing for some reason, this alert will silently stop working. That's bad.

what would be a better approach?

TheoBrigitte:
Replace the `else if` with just an `else`. Better to catch an alert in the wrong team than in none.

fiunchinho:
> Replace the `else if` with just an `else`. Better to catch an alert in the wrong team than in none.

I gave it a shot, please take a look now.

TheoBrigitte (approved):
Looks better thanks

yulianedyalkova (approved):

whites11 (approved):

PR #169, opened 2021-11-30
Add inhibition when cluster has no workers.

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/546

This PR adds a new inhibition alert that fires when a workload cluster has no worker nodes.
This is currently only supported in Azure and AWS as a POC.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add inhibition when cluster has no workers.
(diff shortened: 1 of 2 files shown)


njuettner (approved):
lgtm, so this will create an alert in case of nodes 0. with this we can create general silence for `InhibitionClusterWithoutWorkerNodes`?

fiunchinho (approved):

whites11:
> lgtm, so this will create an alert in case of nodes 0. with this we can create general silence for `InhibitionClusterWithoutWorkerNodes`?

It will work like any other inhibition we have. We will need to tag alerts that will be silenced by this inhibition.

PR #172, opened 2021-12-01
webhook errors for management clusters

njuettner:
I might be wrong, just want to hear some opinion. We only get paged for clusters which are not cluster_type management cluster which is wrong IMHO. We would never be notified in case there's something wrong with kyverno webhook

Example, the metric before changing "!=":
image

Changing it to "=":
image

### Checklist


fiunchinho (approved):
Makes sense

TheoBrigitte (approved):
Sounds alright

PR #181, opened 2021-12-14
fix-alert-missing-aws-admission-manager

calvix:
fix the alert as this label do not exists so it fired everywhere


njuettner (approved):
nice catch

QuentinBisson (approved):

PR #184, opened 2021-12-16
Ownership efk

piontec:
This PR:

- adds per-team ownership of `efk-stack-app` and `nginx-ingress-controller-app` for the `WorkloadClusterAppFailed` alert

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
@@ -124,11 +124,12 @@ spec:
         team: honeybadger
         topic: releng
     ## Workload cluster app rules only available on the management cluster Prometheus because the source is app-exporter
+    ## By default they go to Honeybadger, unless there's an explicit team configured below for specific apps.
     - alert: WorkloadClusterAppFailed
       annotations:
         description: '{{`Workload Cluster App {{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
         opsrecipe: app-failed/
-      expr: label_replace(app_operator_app_info{status!~"(?i:(deployed|cordoned))", catalog!~"control-plane-.*"}, "cluster_id", "$1", "namespace", "(([^g]|g[^i]|gi[^a]|gia[^n]|gian[^t]|giant[^s]|giants[^w]|giantsw[^a]|giantswa[^r]|giantswar[^m])*)") == 1
+      expr: label_replace(app_operator_app_info{status!~"(?i:(deployed|cordoned))", catalog!~"control-plane-.*", app!~"(efk-stack-app)|(nginx-ingress-controller-app)"}, "cluster_id", "$1", "namespace", "(([^g]|g[^i]|gi[^a]|gia[^n]|gian[^t]|giant[^s]|giants[^w]|giantsw[^a]|giantswa[^r]|giantswar[^m])*)") == 1
       for: 30m
       labels:
         area: managedservices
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -140,6 +141,40 @@ spec:
         sig: none
         team: honeybadger
         topic: releng
+    # Alert Atlas about EFK deployment failed as detected by app platform.
+    - alert: WorkloadClusterAppFailed
```
kubasobon, line 145: Can we reuse the alert name? I thought it has to be unique.

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -140,6 +141,40 @@ spec:
...
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "true"
+        severity: page
+        sig: none
+        team: atlas
+        topic: releng
+          # Alert Cabbage about nginx-IC deployment failed as detected by app platform.
+    - alert: WorkloadClusterAppFailed
```
kubasobon, line 162: ditto

helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
```diff
@@ -38,7 +38,7 @@ spec:
     - alert: FluxKustomizationFailed
       annotations:
         description: |-
-          {{`Flux Kustomization on {{ $labels.installation }}/{{ $labels.cluster_id }} in {{ $labels.namespace }}/{{ $labels.name }} is stuck in Failed state.`}}
+          {{`Flux Kustomization {{ $labels.name }} in ns {{ $labels.namespace }} on {{ $labels.installation }}/{{ $labels.cluster_id }} is stuck in Failed state.`}}
```
kubasobon, line 41: Can you please do the same for `FluxHelmReleaseFailed` and `FluxSourceFailed`?

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
piontec: MIssed that, thanks!

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -140,6 +141,40 @@ spec:
         sig: none
         team: honeybadger
         topic: releng
+    # Alert Atlas about EFK deployment failed as detected by app platform.
```
QuentinBisson, line 144: Is it not possible to use the team annotation set in the chart instead ? cc @JosephSalisbury

rossf7: Routing by app name will work for a few apps but yes we could use the team label instead.

The [WorkloadClusterAppFailed](https://github.com/giantswarm/prometheus-rules/blob/4e7df4887b8b36a9f8af436c6bf5a4fcf37bbe55/helm/prometheus-rules/templates/alerting-rules/app.rules.yml#L127-L142) alert needs to be split by team like we do for MCs. e.g. [ManagementClusterAppFailedAtlas](https://github.com/giantswarm/prometheus-rules/blob/4e7df4887b8b36a9f8af436c6bf5a4fcf37bbe55/helm/prometheus-rules/templates/alerting-rules/app.rules.yml#L14-L29)

We just need to be careful with the statuses. I would just route `failed` apps for now. With other statuses like `not-installed` if chart-operator is broken we could page multiple teams which would be dumb

kubasobon (approved):

PR #197, opened 2022-01-03
Fix description for alert

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/664

This alert sometimes does not contain the name of the critical pod that triggered the alert in the description:

Screenshot_20220103_171548

I noticed this happens when the `absent(...)` query is triggering the alert.
In fact, the `absent` query result only contains the `container` label, but we use the `pod` label to populate the description.

This PR adds a few `label_replace` clauses to ensure the `pod` label is always set and thus the description contains useful data.


fiunchinho (approved):
nice!!

PR #217, opened 2022-01-19
Push to openstack app collection

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/20430

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ workflows:
     jobs:
       - architect/push-to-app-catalog:
           context: "architect"
-          name: app-catalog
+          name: push-to-app-catalog
           app_catalog: "control-plane-catalog"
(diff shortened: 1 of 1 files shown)


kopiczko (approved):

QuentinBisson (approved):

TheoBrigitte (changes requested):
Ok, but could we not add this `push-to` prefix everywhere please ? I find that it makes things less readable.

tfussell:
> Ok, but could we not add this `push-to` prefix everywhere please ? I find that it makes things less readable.

@TheoBrigitte I did this because we use this naming convention in basically all of our other CI pipelines and I think the consistency is important. I found it initially confusing when I looked at the file and the job name didn't contain `push-to-`. Maybe something to bring to SIG Releng?

tfussell:
Closing in favor of https://github.com/giantswarm/prometheus-rules/pull/214

PR #237, opened 2022-01-27
add `PrometheusRuleFailures`

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/20510

This PR adds the `PrometheusRuleFailures` alert which will notify us when prometheus rules evaluation fails. This is bad because if prometheus can't evaluate a rule we will never get alerted for it.

i.e. currently failing SLO rules is failing and went unoticed

```
msg="Evaluating rule failed" rule="alert: ServiceLevelBurnRateTooHigh ...
```

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
@@ -43,6 +43,7 @@ spec:
         description: {{`Prometheus {{$labels.installation}}/{{$labels.cluster_id}} has failed to evaluate rule(s) {{ printf "%.0f" $value }} time(s).`}}
         summary: Prometheus is failing rule evaluations.
       expr: increase(prometheus_rule_evaluation_failures_total[30m]) > 0
+      for: 1h
       labels:
         area: empowerment
         severity: page
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
@@ -42,7 +42,7 @@ spec:
       annotations:
         description: {{`Prometheus {{$labels.installation}}/{{$labels.cluster_id}} has failed to evaluate rule(s) {{ printf "%.0f" $value }} time(s).`}}
         summary: Prometheus is failing rule evaluations.
-      expr: increase(prometheus_rule_evaluation_failures_total[30m]) > 0
+      expr: rate(prometheus_rule_evaluation_failures_total[30m]) > 0
       for: 1h
       labels:
         area: empowerment
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
@@ -38,3 +38,11 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: PrometheusRuleFailures
+      annotations:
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
```diff
@@ -38,3 +38,10 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: PrometheusRuleFailures
+      annotations:
+        description: {{`Prometheus {{$labels.namespace}}/{{$labels.pod}} has failed to evaluate {{ printf "%.0f" $value }} rules.`}}
+        summary: Prometheus is failing rule evaluations.
+      expr: increase(prometheus_rule_evaluation_failures_total[30m]) > 0
+      labels:
+        severity: critical
```
JosephSalisbury, line 47: what does `critical` mean?

TheoBrigitte: eh sorry, that was a copy paste

JosephSalisbury: I could believe only having this in business hours would be nice too

TheoBrigitte:
Tested and looks alright :heavy_check_mark: 

image

Could not extract the rule group name from labels. And adding something like the following to the description feels a bit much.

```
/etc/prometheus/rules/prometheus-ghost-rulefiles-0/monitoring-service-level.rules-04ac4d70-77e5-4d2c-8e1a-5ee85bd9f529.yaml;service-level
```

helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
TheoBrigitte: Adjusted labels

helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
```diff
@@ -38,3 +38,15 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: PrometheusRuleFailures
+      annotations:
+        description: {{`Prometheus {{$labels.installation}}/{{$labels.cluster_id}} has failed to evaluate rule(s) {{ printf "%.0f" $value }} time(s).`}}
+        summary: Prometheus is failing rule evaluations.
+      expr: increase(prometheus_rule_evaluation_failures_total[30m]) > 0
```
JosephSalisbury, line 45: why `increase` over `rate`

TheoBrigitte: Yeah good point `rate` should be used an not `increase`.

helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
```diff
@@ -38,3 +38,15 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: PrometheusRuleFailures
+      annotations:
+        description: {{`Prometheus {{$labels.installation}}/{{$labels.cluster_id}} has failed to evaluate rule(s) {{ printf "%.0f" $value }} time(s).`}}
```
JosephSalisbury, line 43:
```suggestion
        description: {{`Prometheus has failed to evaluate rule(s) {{ printf "%.0f" $value }} time(s).`}}
```

JosephSalisbury: we already have this info in the title

JosephSalisbury (approved):
small nit, but lgtm otherwise

PR #249, opened 2022-02-03
add managed prometheus to ServiceLevelBurnRateTooHigh

TheoBrigitte:


TheoBrigitte:
Tested and working on `godsmack`

image

helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
```diff
@@ -287,6 +287,26 @@ spec:
         provider: aws
       record: slo_target
 
+    # -- Managed Prometheus
+    - expr: (up{app="prometheus-operator-app-prometheus",container="prometheus"}*0)+1
```
webwurst, line 291: What does this mean `*0`?

TheoBrigitte: Added some comments to explain better what we're doing.

But basically we ignore the value reported by `up` and always set it to 1 when a managed Prometheus is present. `*0` is basically ignoring the value then `+1` set it to always be `1`.

PR #250, opened 2022-02-07
fix slo

TheoBrigitte:
- remove cluster_id from slo_target
- remove provider label from SLOs
- restore group_left

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/service-level.rules.yml
@@ -16,15 +16,15 @@ spec:
         opsrecipe: service-level-burn-rate-too-high/
       expr: |
         (
-          slo_errors_per_request:ratio_rate1h{service!~"efk-.*|.*external-dns.*|kong-.*|.*nginx-.*"} > on (service) group_right slo_threshold_high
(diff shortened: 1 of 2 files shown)


TheoBrigitte:
:heavy_check_mark:  tested with blackbox exporter on puma

#### Before
image

#### After
image

TheoBrigitte:
:white_check_mark: tested kubelet on `gauss`

image

SLO for kubelet itself never paged after killing several nodes on a 3-5 worker cluster. Feels like something is wrong but the SLO rules were executed correctly.

QuentinBisson:
Can you explain why you removed the cluster_id ? Because it looks like we are doing SLO across clusters in one installation

TheoBrigitte:
> Can you explain why you removed the cluster_id ? Because it looks like we are doing SLO across clusters in one installation

I did that otherwise we get multiple entries for each `service` on SLO targets. AFAIK we only define SLO targets per app/service and not per cluster.

QuentinBisson (approved):

TheoBrigitte:
> well, you're right :) Also we have separate proemtheuses per cluster so my reasoning does not stand :)

All good, the SLO setup is quite complex. BTW I experimented with [sloth](https://github.com/slok/sloth) to try and replace our current SLO setup but it felt tedious to convert all our current SLO rules.

PR #263, opened 2022-02-16
Raise timeout before alerting for alerts that have `cancel_if_kube_state_metrics_down`.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/20888

This PR:

Raises alert timeout for alerts that have cancel_if_kube_state_metrics_down = true.
This is because when a customer deletes all node pools from a cluster, kube_state_metrics goes down almost immediately but the inhibition might take a while before is in place. Increasing the timeout should help with false positives.

### Checklist


njuettner (approved):
25m is a arbitrary number right?

PR #292, opened 2022-03-18
Inhibit openstack alerts out of business hours

pipo02mix:
Towards https://github.com/giantswarm/roadmap/issues/908

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/_helpers.tpl
+++ b/helm/prometheus-rules/templates/_helpers.tpl
@@ -34,3 +34,11 @@ rocket
 phoenix
 {{- end -}}
 {{- end -}}
+
+{{- define "hasToBeCancelled" -}}
+{{- if has .Values.managementCluster.provider.kind (list "openstack") -}}
+"true"
+{{- else -}}
+"false"
+{{- end -}}
+{{- end -}}
--- a/helm/prometheus-rules/templates/_helpers.tpl
+++ b/helm/prometheus-rules/templates/_helpers.tpl
@@ -35,7 +35,7 @@ phoenix
 {{- end -}}
 {{- end -}}
 
-{{- define "hasToBeCancelled" -}}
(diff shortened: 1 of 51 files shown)


kopiczko (approved):
I like it

helm/prometheus-rules/templates/_helpers.tpl
```diff
@@ -34,3 +34,11 @@ rocket
 phoenix
 {{- end -}}
 {{- end -}}
+
+{{- define "hasToBeCancelled" -}}
```
QuentinBisson, line 38: I would probably rename this field to ensure this describe things best. Why are you cancelling alert outside of working hours ?

pipo02mix: because CAPO is not yet prod ready

QuentinBisson: Yes that's what I was thimking :D Could we call that beta or testing instead ?

QuentinBisson: isBeta for example ?

whites11 (approved):
I dont see problems from my point of view

helm/prometheus-rules/templates/_helpers.tpl
pipo02mix: but my thinking it could have other reasons why we could inhibit them. @kopiczko what do you think? you are good at naming

QuentinBisson: Makes sense yes :)

kopiczko: > @kopiczko what do you think?

I don't mind. If this is the only reason for this thing to exist then probably changing the name makes sense.

QuentinBisson (approved):

QuentinBisson:
<3

PR #296, opened 2022-03-23
[MC] Add new alert for when `Kyverno` scraping fails for 5m.

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/903

@giantswarm/team-honeybadger what do you think about this?
If anything happens to a management cluster and it turns down all replicas of kyverno, the cluster stops working until kyverno runs again.
As owners of kyverno, reliability is your responsibility as much as reliability of the cluster nodes health is Phoenix's.

We therefore believe that if such thing happens both us and you should be paged.

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/kyverno.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/kyverno.rules.yml
@@ -1,0 +1,26 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  creationTimestamp: null
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+    cluster_type: "management_cluster"
+  name: kyverno.management_cluster.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+    - name: kyverno
+      rules:
+        - alert: NoKyvernoPodRunning
+          annotations:
+            description: '{{`Scraping of all kyverno pods failed for 5 minutes.`}}'
+            opsrecipe: kyverno-is-not-running/
+          expr: (count(up{service="kyverno-svc-metrics"}) or vector(0)) == 0
+          for: 5m
+          labels:
+            area: kaas
+            severity: page
+            team: honeybadger
+            topic: managementcluster
+            cancel_if_cluster_status_creating: "true"
+            cancel_if_cluster_status_deleting: "true"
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/kyverno.rules.yml
```diff
@@ -0,0 +1,26 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  creationTimestamp: null
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+    cluster_type: "management_cluster"
+  name: kyverno.management_cluster.rules
```
MarcelMue, line 8: Why not `kyverno.rules`? Or do we have this naming anywhere else? I thought it was just the filename

helm/prometheus-rules/templates/alerting-rules/kyverno.rules.yml
```diff
@@ -0,0 +1,26 @@
...
+            opsrecipe: kyverno-is-not-running/
+          expr: (count(up{service="kyverno-svc-metrics"}) or vector(0)) == 0
+          for: 5m
+          labels:
+            area: kaas
+            severity: page
+            team: honeybadger
+            topic: managementcluster
+            cancel_if_cluster_status_creating: "true"
+            cancel_if_cluster_status_deleting: "true"
```
AverageMarcus, line 26: We should prob also add `cancel_if_cluster_status_updating` here.

MarcelMue (commented):
I am generally not against this - let's see how often it actually pages?

helm/prometheus-rules/templates/alerting-rules/kyverno.rules.yml
whites11: no, that's the whole point. Kyverno has to survive during a cluster upgrade

helm/prometheus-rules/templates/alerting-rules/kyverno.rules.yml
whites11: honestly I don't know what that `name` field does at all. I followed the structure I found here https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/alerting-rules/kvm.management-cluster.rules.yml#L9

MarcelMue: The only thing I see is, that it matches the file name - so `kyverno.rules` seems correct to me.

whites11: changed

MarcelMue (approved):
Ok for me to try this and see how badly it explodes.

PR #310, opened 2022-04-08
add PrometheusTargetMissing alert

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/21156

Add `PrometheusTargetMissing` alert, so we are alerted when a job has 0 targets.

This needs to be tested everywhere first, because I have the feeling many targets are missing.


TheoBrigitte:
31 targets missing on `gamma`

image

21 targets missing on `gauss`

image

TheoBrigitte:
This addressed as part of https://github.com/giantswarm/giantswarm/issues/24062 we now have a `PrometheusJobScrapingFailure` alert and this is being continuously improved.

PR #314, opened 2022-04-11
Disable vault and cluster-service rules for CAPO and CAPV

erkanerol:
Towards https://github.com/giantswarm/roadmap/issues/640

We don't install `cluster-service` into CAPO and CAPV clusters. We don't have `vault` machine as well. We deploy vault into MC now. Vault monitoring will be supported soon but we need to disable those now.

### Checklist


QuentinBisson (approved):
LGTM if tested

erkanerol:
@QuentinBisson I tested by setting `managementCluster.provider.kind` to `kvm` and `openstack`.

PR #323, opened 2022-04-20
Nodepool SLO alert

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/21428

The current SLO alert for the `kubelet` service takes into account all nodes as a whole, not considering single node pools.
Sometimes this might be not ideal, because one node being down in a 10 nodes node pool as a different impact than a node being down in a 2 nodes node pool.

Starting from kube-state-metrics-app version 1.10.0, we now have the `nodepool` label available so we can split node health metrics by node pool.

This PR leverages this new available info and exposes SLO metrics by node pool.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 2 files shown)


QuentinBisson (approved):

anvddriesch (approved):
I haven't tested it but it looks good.

PR #331, opened 2022-05-04
Rename alert

whites11:
Rename `ClusterCertificateWillExpireMetricMissing` alert to `ClusterCertificateExpirationMetricsMissing` to avoid being paged for test installations.

After turning ClusterCertificateWillExpireMetricMissing from a Notify to a Page alert, it started paging for all clusters, including testing ones.
The reason is that, in the notification policy, we page for test clusters if alerts contain the word `CertificateWillExpire` in the `message`:

test

This PR renames the alert in order not to contain that word in the message.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Rename `ClusterCertificateWillExpireMetricMissing` alert to `ClusterCertificateExpirationMetricsMissing` to avoid being paged for test installations.
+
 ## [2.16.0] - 2022-05-04
 
 ### Added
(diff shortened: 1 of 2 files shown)


zewolfe (approved):
LGTM 👍🏾

QuentinBisson:
That looks like a bad reason to rename imo

QuentinBisson:
Why do we cancel the alerts starting with CertificateWillExpire ? What's the use of having them if we ignore them anyway ?

whites11:
> Why do we cancel the alerts starting with CertificateWillExpire ? What's the use of having them if we ignore them anyway ?

that's the opposite. The alert rule always fires if the name matches. But we want this specific alert to NOT fire for test installations because it makes no sense at all.

QuentinBisson:
My apologies ;)

QuentinBisson (approved):

PR #333, opened 2022-05-06
Fix service level rule failing for prometheus

QuentinBisson:
This PR:

fixes prometheus rule failing due to 2 slo target for managed prometheus as the slo target is also created by  

- expr: sum by (service, area) (raw_slo_errors{area="managed-apps"} - raw_slo_errors{area="managed-apps"}) + 1-0.99
  record: slo_target

image
image

### Checklist


TheoBrigitte (approved):
Fine when tested.

PR #337, opened 2022-05-11
add irsa-operator as a new alert for phoenix

njuettner:
This PR:

- adds/changes/removes etc

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/aws.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/aws.management-cluster.rules.yml
@@ -148,7 +148,7 @@ spec:
         topic: kubernetes
     - alert: IRSATooManyErros
       annotations:
-        description: '{{`IRSA bootstrap for Cluser {{ $labels.cluster_id }} failed.`}}'
+        description: '{{`IRSA bootstrap for Cluster {{ $labels.cluster_id }} failed.`}}'
         opsrecipe: irsa-operator-error/
       expr: irsa_operator_cluster_errors > 0
       for: 10m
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/aws.management-cluster.rules.yml
```diff
@@ -146,4 +146,16 @@ spec:
         severity: page
         team: phoenix
         topic: kubernetes
+    - alert: IRSATooManyErros
+      annotations:
+        description: '{{`IRSA bootstrap for Cluster {{ $labels.cluster_id }} failed.`}}'
+        opsrecipe: irsa-operator-error/
+      expr: irsa_operator_cluster_errors > 0
```
whites11, line 153: 0 seems to me a very low threshold.
Also, would it be possible to make it a SLO-based alert?

njuettner: The thing is whenever there's one error in the chain there is something wrong and it will not recover. 

My intention was to make it really noisy in the beginning so we can probably smash all errors in case there are some or we immediately fix for instance the permissions if they are not set right and later we turn it into something SLO like because we have more experience and it will unlikely that there are bigger issues

TheoBrigitte: Fine with me when that's a conscious choice.

TheoBrigitte (approved):

helm/prometheus-rules/templates/alerting-rules/aws.management-cluster.rules.yml
whites11: okay

whites11 (approved):

PR #339, opened 2022-05-12
Improve '*EtcdDBSizeTooLarge' alerts.

whites11:
What defines an etcd as "too large"? I reckon not the absolute size because a 100 nodes cluster will definitely have a bigger etcd database compared to a 10 nodes one.
In my opinion, this alert should fire then the total size on disk of etcd DB is bigger than the used size as reported by etcd metrics.
That means that the database needs a defragmentation to free disk space (and that's what the ops recipe wants to do).

This PR changes the query, in order for the alert not to page when the absolute size is > 500M, but to page when the difference between disk and used space differ by more than 200M.

WDYT?

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Improve '*EtcdDBSizeTooLarge' alerts.
+
 ## [2.18.0] - 2022-05-11
 
 ### Added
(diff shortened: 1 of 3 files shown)


QuentinBisson:
Just a small question, especially with CAPI, would it not make sense to defragment the etcd containers with a cron job on a regular schedule ?

whites11:
> Just a small question, especially with CAPI, would it not make sense to defragment the etcd containers with a cron job on a regular schedule ?

that happens already with cron and etcd backup operator does that automatically before each upgrade (every 6h)

PR #348, opened 2022-05-17
add support for vcd provider"

glitchcrab:
towards https://github.com/giantswarm/giantswarm/issues/22106

- add support for VCD provider

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/_helpers.tpl
+++ b/helm/prometheus-rules/templates/_helpers.tpl
@@ -28,7 +28,7 @@ giantswarm.io/service-type: {{ .Values.serviceType }}
 {{- end -}}
 
 {{- define "providerTeam" -}}
-{{- if has .Values.managementCluster.provider.kind (list "kvm" "vsphere" "openstack") -}}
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/_helpers.tpl
```diff
@@ -44,31 +44,31 @@ phoenix
...
+{{- if has .Values.managementCluster.provider.kind (list "openstack" "vcd" "vsphere") -}}
 false
 {{- else -}}
 true
 {{- end -}}
 {{- end -}}
 
 {{- define "isBastionBeingMonitored" -}}
-{{- if has .Values.managementCluster.provider.kind (list "openstack" "vsphere") -}}
+{{- if has .Values.managementCluster.provider.kind (list "openstack" "vcd" "vsphere") -}}
```
kopiczko, line 71: Possibly this should be based on negation at this point

kopiczko (approved):

helm/prometheus-rules/templates/_helpers.tpl
QuentinBisson: Should we maybe create a isOnPremise property up the config chain instead ?  Or maybe something like location: cloud || on premise and that property is calculated from the provider ?

PR #355, opened 2022-05-26
Split mixin file and add script to sync it

mcbenjemaa:
This PR:

- Split mixin file 
- add script to sync rules
- change the notes of mixin update

issue: https://github.com/giantswarm/giantswarm/issues/19855

### Checklist

- [X] Consider creating a dashboard (if it does not exist already) to help oncallers monitor the status of the issue.

### DIFF
--- a/scripts/sync-kube-mixin.sh
+++ b/scripts/sync-kube-mixin.sh
@@ -26,12 +26,9 @@ cd "${TMPDIR}"/mixins
 MIXIN_VER=$(git rev-parse HEAD)
 cd - > /dev/null
 
+RULESFILE="helm/prometheus-rules/templates/recording-rules/kubernetes-mixins.rules.yml"
 
-#CONTENT=$(cat "${TMPDIR}"/mixins/files/prometheus-rules/rules.yml| yq -P .groups)
-
-cp "${TMPDIR}"/mixins/files/prometheus-rules/rules.yml helm/prometheus-rules/templates/recording-rules/kubernetes-mixin-test.yml
-
-CONTENT='apiVersion: monitoring.coreos.com/v1
+PRECONTENT='apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
   labels:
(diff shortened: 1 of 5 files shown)


hervenicol:
Should we wait for https://github.com/giantswarm/dashboards/pull/124 to be merged before cleaning the doc?

mcbenjemaa:
@hervenicol  Yes 👍

mcbenjemaa:
@QuentinBisson We agreed to split the mixin rules file,

However, Do we need to add a script that syncs rules, the same as we did in dashboards?

mcbenjemaa:
@QuentinBisson We agreed to split the mixin rules file,

However, Do we need to add a script that syncs rules, the same as we did in dashboards?

QuentinBisson:
We will need to split the kube-prometheus-rules from the mixin files though

mcbenjemaa:
@QuentinBisson  mixin rules are split and a script has been added to sync the rules.
can you take a look.

scripts/sync-kube-mixin.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# make a temporary dir to work in
+TMPDIR=$(mktemp -d -t 'tmp.XXXXXXXXXX')
```
hervenicol, line 8: Missing quotes. Did you use `shellcheck`? ;)
There's a few other vars without quotes in the script.

scripts/sync-kube-mixin.sh
```diff
@@ -0,0 +1,50 @@
...
+trap 'rm -f $TMPDIR' EXIT
```
hervenicol, line 9: Do you want to `rm -f` or `rm -rf` at exit, like at the end of the script?
Maybe that should be factorized in a common function?

scripts/sync-kube-mixin.sh
```diff
@@ -0,0 +1,50 @@
...
+MIXIN_REPO="git@github.com:giantswarm/giantswarm-kubernetes-mixin.git"
+# clone a branch of tag if provided
+BRANCH=${1:-""}
+
+if [ -z "${BRANCH}" ]; then
```
hervenicol, line 14: I'm not a big fan of curly braces when not needed. I don't think it improves readability, it's just extra characters.
But it's my personal opinion / nitpicking ;)

scripts/sync-kube-mixin.sh
```diff
@@ -0,0 +1,50 @@
...
+  name: kube-mixins.recording.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+'
+
+# copy generated rules file
+cp "${TMPDIR}"/mixins/files/prometheus-rules/rules.yml ${RULESFILE}
+
+# prepend K8s objectmeta to the rules file
+printf '%s  %s' "${PRECONTENT}" "$(cat ${RULESFILE})" >${RULESFILE}
```
hervenicol, line 45: Is this supposed to:
- write `$PRECONTENT`
- write `$RULESFILE` contents by adding an indentation at the beginning of each line?

If so, I don't understand how it works, and it does not seem to work on my laptop with bash.

On my laptop, here is what it does:
- write `$PRECONTENT`
- write `$RULESFILE` contents by adding an indentation at **the first** line

hervenicol (changes requested):

scripts/sync-kube-mixin.sh
mcbenjemaa: I LIKE double braces in if conditions, I can distinguish that it's an if statement :D

scripts/sync-kube-mixin.sh
hervenicol: I mean, same code for same usage in 2 places, I thought it deserves a function. The syntax difference just proved the point :grin:

scripts/sync-kube-mixin.sh
mcbenjemaa: You suggest omitting curly braces `{ }` from each variable?

hervenicol: Yes, when not needed. Just a suggestion, nothing critical there: we have no coding style guidelines for bash ;)

hervenicol (approved):
All good! :+1:

PR #359, opened 2022-05-31
add openstack lb alert

bavarianbidi:
This PR:

- adds kubernetes API down alert on openstack

Ops-recipe: https://github.com/giantswarm/giantswarm/pull/22258

Fixes: https://github.com/giantswarm/giantswarm/issues/22251

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Introduce Kubernetes API down alert on openstack.
+
 ## [2.25.0] - 2022-05-26
 
 ### Fixed
@@ -171,7 +175,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Removed `ManagementClusterPodPendingFor15Min` and `ManagementClusterPodPending`.
-
 
 ## [1.7.1] - 2022-03-22
 
@@ -437,7 +440,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.40.2] - 2021-12-03
 
-
 ### Fixed
 
 - Consolidate alerts that covered the same issues.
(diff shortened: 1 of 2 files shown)


QuentinBisson (approved):

bavarianbidi:
> Is this alert not already present ?

i only found some recording rules for SLOs, but no active alerting rule. We had an issue last friday where k8s API wasn't reachable and no alert triggered us, therefore we decided to implement this alert

QuentinBisson (approved):
I agree this makes sense.
 I'm wondering whether openstack could use the blackbox exporter as well (that's an exporter used to make network connections so we could ensure the api server is reachable from outside the management clusrter also if that is needed)

bavarianbidi:
> I agree this makes sense. I'm wondering whether openstack could use the blackbox exporter as well (that's an exporter used to make network connections so we could ensure the api server is reachable from outside the management clusrter also if that is needed)

:-) yep, that would be cool (created for myself some notes this morning @QuentinBisson FYI: xref: https://github.com/giantswarm/giantswarm/issues/22251#issuecomment-1141687586)

QuentinBisson:
That would be cool :)

pipo02mix (approved):
LGTM

fiunchinho:
Just my two cents. Back in the day we started monitoring our clusters following this approach, but then we moved towards SLO based alerts. Maybe there are good reasons to use this alert the way it has been written, but it feels a step backwards. That's why we have the SLO recording rule for the api server. What do you think @giantswarm/team-atlas ?

fiunchinho:
Just my two cents. Back in the day we started monitoring our clusters following this approach, but then we moved towards SLO based alerts. Maybe there are good reasons to use this alert the way it has been written, but it feels a step backwards. That's why we have the SLO recording rule for the api server. What do you think @giantswarm/team-atlas ?

QuentinBisson:
I agree with what you are saying @fiunchinho but from what I understood, this alert test that the api server is accessible from the management cluster (i.e. connectivity).
This might be covered by one of our prometheus alerts though so the blackbox exporter might be a better option for this

bavarianbidi:
DO NOT MERGE (PR might be closed without merging it - xref: https://github.com/giantswarm/giantswarm/issues/22251#issuecomment-1143731377)

PR #362, opened 2022-06-02
Use rate on sum to avoid issues when resetting

paurosello:
This PR:

Change from sun of requests to rate of requests 

Related to https://github.com/giantswarm/giantswarm/issues/22134

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixing mixins prometheus recording rules
+- Use rate instead of the total amount of requests for calculating SLO for kube-scheduler
 
 ## [2.24.0] - 2022-05-25
 
--- a/helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
+++ b/helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
@@ -252,57 +252,57 @@ spec:
 
     # core k8s components internal API requests
     # record number of requests.
-    - expr: label_replace(sum(rest_client_requests_total{app=~"kube-controller-manager|kube-scheduler"}) by (cluster_id,app), "service", "$1", "app", "(.*)")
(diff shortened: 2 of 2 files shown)


paurosello:
The problem is I don't know how to test this :(

whites11 (approved):
At a glance it makes sense, but I know SLO alerts calculate a rate on their own so it's important Atlas reviews this as well

hervenicol:
:exclamation: Changing a (broken) counter to a gauge is a breaking change for anything which relies on these metrics.

You can't sum counters in a reliable way, that's a known problem with Prometheus, so I think you're doing right there. :+1: 

Just, maybe give the metrics a new name? Downside is that may create some unused metrics, and use space for nothing.
But do we have a reliable way to know which metrics are used and where? Maybe a few greps in the right repos would be enough to be safe on that?

Let my Atlas colleagues who know better than me complete my comment :wink:

paurosello:
> ❗ Changing a (broken) counter to a gauge is a breaking change for anything which relies on these metrics.
> 
> You can't sum counters in a reliable way, that's a known problem with Prometheus, so I think you're doing right there. 👍
> 
> Just, maybe give the metrics a new name? Downside is that may create some unused metrics, and use space for nothing. But do we have a reliable way to know which metrics are used and where? Maybe a few greps in the right repos would be enough to be safe on that?
> 
> Let my Atlas colleagues who know better than me complete my comment 😉

I am only modifying a subset of the records included in `raw_slo_requests`, so I don't think we can't really rename without creating a mess as all the alarms are based on this metric.

hervenicol:
> I am only modifying a subset of the records included in `raw_slo_requests`, so I don't think we can't really rename without creating a mess as all the alarms are based on this metric.

If I understood correctly, this PR changes **counters** (a metric that only increases, and counts the amount "since ever") into **gauges** (a metric that can gives an instant value, and can be totally different from one read to another).

These are totally different metrics, and changing the contents of the metrics without changing its name will still create a mess with all that's based on it IMO.

paurosello:
Reverted as we are not sure if the alerts still work.

@JosephSalisbury do you have an opinion on this? As you are the SLO master... :)

JosephSalisbury:
@paurosello i'd have to pair with you on this to get into context, feel free to drop an hour in my calendar sometime and we can smash it out <3

PR #363, opened 2022-06-03
Enable Monitoring during upgrade

paurosello:
This PR:

- Enable many alerts that were deactivated during upgrades.

The main reason behind this PR is that we have faced multiple issues that should have alerted us during upgrades but were silenced. As we are moving towards the automated upgrades we should strive to improve our components resiliency in order to achieve seamless upgrades.

An example of silenced alerts during upgrades that should have paged us:
- All masters were down, API was down and we did not get paged.
- CoreDNS is down we also don't get paged

### Checklist


QuentinBisson (approved):
LGTM

TheoBrigitte (approved):

PR #368, opened 2022-06-09
added rules for Loki

hervenicol:
This PR adds alerts for Loki

Ops-recipes: https://github.com/giantswarm/giantswarm/pull/22375
Towards https://github.com/giantswarm/roadmap/issues/718

No dashboards created yet. I hope to use loki-mixins in the future, but we should have some working loki instances first!

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Alerts for Loki
+
 ## [2.26.0] - 2022-06-08
 
 ### Added
--- a/helm/prometheus-rules/templates/alerting-rules/loki.all.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/loki.all.rules.yml
@@ -1,0 +1,50 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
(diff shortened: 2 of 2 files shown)


QuentinBisson (approved):
LGTM. As you said, once we get the mixin running maybe we can also get the alerts from them

QuentinBisson:
What do you think about adding up alerts for hte loki components as well ? Unless we want to start on a loose SLO ?

TheoBrigitte (commented):
I feel like those alerts are quite sensitive. Could you please increase the time windows? To hours maybe?

hervenicol:
@QuentinBisson :
> What do you think about adding up alerts for the loki components as well ? Unless we want to start on a loose SLO ?

The only missing component there is the compactor (the only one that does not manage actual requests), and the official mixins use data from a rewriterule for that. So I propose to add it later, when we have rewriterules in place.

If you are talking about checking an instance of Loki, with all its components, is currently running, that has to be optional for the moment.

I'm not sure I see what you have in mind, maybe you could give more details?
Anyway, I feel like today we lack knowledge to create meaningful alerts.

@TheoBrigitte :
> I feel like those alerts are quite sensitive. Could you please increase the time windows? To hours maybe?
* for error percentage, why not; done :heavy_check_mark: 
* for panic messages though, I think if they set it this way it means only 1 panic message means there is a critical problem, and there may be no more messages after that. I'd prefer we wait and have some real-life feedback before we change this one.

QuentinBisson (approved):
LGTM for now as those are not paging

PR #384, opened 2022-07-05
Initial rule test support

bavarianbidi:
This PR:

- adds support for unit testing alerting rules - needed as i've added my first bunch of new alerts

TODO:
- [x] integrate into ci
- [x] finalize make targets
- [x] finish test script
- [x] documentation

### Checklist


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add Rules for Cluster API and Cluster API Openstack for team rocket
```
QuentinBisson, line 14: Can you add something about the testing mechanism for rules ? Or maybe split it up in another PR afterward :)

helm/prometheus-rules/.helmignore
```diff
@@ -0,0 +1,3 @@
+.gitignore
+alerting-rules-test/*.yml
+alerting-rules-test/rules/*.yml
```
QuentinBisson, line 3:
```suggestion
alerting-rules-test/rules/*.yml

```

test/hack/bin/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Ignore everything in this directory
+*
+# Except the following
+!.gitignore 
+!verify-rules.sh
+!fetch-tools.sh
```
QuentinBisson, line 6:
```suggestion
!fetch-tools.sh

```

QuentinBisson:
I realllllly like your approach here <3

Could you write a bit of documentation in the end so we can announce that in news-ops so other team starts to test their alerts also ?

bavarianbidi:
> I realllllly like your approach here <3
> 
> Could you write a bit of documentation in the end so we can announce that in news-ops so other team starts to test their alerts also ? 

Sure, will do that ... Initial intention was to get some early feedback if it make sense to continue on this.

Will write some documentation later on

hervenicol:
I've had a look; ok maybe more than a look, so here is a PR to help things forward: https://github.com/giantswarm/prometheus-rules/pull/417

bavarianbidi:
THX @hervenicol for your PR on this branch :heart: 

I changed the directory structure a bit and add some initial notes about testing and how to add new tests.
For validation i added test cases for all `CAPI` and `CAPO` rules (to have a valid test target).

There are still some limitations (described in `README.md`) but maybe we can start with this and improve the script when we are at the point where we need the improvements.

README.md
```diff
@@ -7,6 +7,47 @@
...
+### Testing
+
+By creating unit tests for Alerting rules it's possible to get early feedback about possible misbehavior in alerting rules.
+Unit tests are executed via `promtool` (part of `prometheus`).
+
+By running `make test` in your local environment, all required binaries will be downloaded and tests will be executed.
+
+#### Writing new tests
+
+1. remove the rules file you would like to test from `test/conf/promtool_ignore`
```
hervenicol, line 19: Should be the other way around IMO.
The current behaviour for each rule is to:
1- check - does not need any test file
2- test - skipped if there is no test file

So why not check all rule files by default?
In my tests, I only need an ignore entry for rules that helm complains about with:
```
Error: could not find template templates/alerting-rules/aws.management-cluster.rules.yml in chart
```

BTW, this makes me wonder if the list of ignored rules may be different for each provider ?

README.md
```diff
@@ -7,6 +7,47 @@
...
+1. create a new test file in [unit testing rules] format for each provider in `test/conf/providers` in `test/tests/providers/<providers>/`
+1. by running `make test` you can validate the your testing rules.
```
hervenicol, line 21:
```suggestion
1. by running `make test` you can validate your testing rules.
```

test/hack/bin/verify-rules.sh
```diff
@@ -0,0 +1,134 @@
...
+
+            echo "###    promtool check rules $GIT_WORKDIR/test/tests/providers/$provider/$filename"
+            local promtool_check_output
+            promtool_check_output="$("$GIT_WORKDIR/$PROMTOOL" check rules "$GIT_WORKDIR/test/tests/providers/$provider/$filename" 2>&1)" ||
+                promtool_check_errors+=("$promtool_check_output")
+
+            local promtool_test_output
+            local testfile="$GIT_WORKDIR/test/tests/providers/$provider/${filename%.yml}.test.yml"
+            if [[ ! -f "$testfile" ]]; then
+                echo "###    testfile $testfile not found, skipping promtool test"
```
hervenicol, line 98: Bug: It says it skips, then it does not.
```suggestion
                echo "###    testfile $testfile not found, skipping promtool test"
                continue
```

hervenicol (changes requested):

hervenicol:
Totally agree we should not wait to have something perfect before we merge, and the current status is already great!
I'd just like to clarify stuff around the `promtool_ignore` file before.

hervenicol:
Status:
* we need provider-specific ignore files and provider-specific tests
* bust most of the tests and ignored rules are generic
* So we need to have a kind of hierarchy for them.

Let's have the right structure before we merge, so we do things clean from the start when people start contributing.

test/hack/bin/verify-rules.sh
```diff
@@ -0,0 +1,166 @@
...
+            "$GIT_WORKDIR/$HELM" template \
+                --set="managementCluster.provider.kind=$provider" \
+                --release-name prometheus-rules \
+                --namespace giantswarm "$GIT_WORKDIR"/helm/prometheus-rules \
+                -s "$file" |
+                "$GIT_WORKDIR/$YQ" '.spec' - >"$GIT_WORKDIR/test/tests/providers/$provider/$filename"
+
+            echo "###    promtool check rules $GIT_WORKDIR/test/tests/providers/$provider/$filename"
+            local promtool_check_output
+            promtool_check_output="$("$GIT_WORKDIR/$PROMTOOL" check rules "$GIT_WORKDIR/test/tests/providers/$provider/$filename" 2>&1)" ||
```
bavarianbidi, line 107: to be more strict, we could think about doing the `promtool check rules` command (validates the syntax of the rules) for every rules file and doing the unit-testing later on.

But maybe this is hopefully a motivation for everyone getting rule file validation by implementing unit tests :sweat_smile:

hervenicol: I thought about that, but that means having another set of ignore files, because some rule files can't even be extracted from helm, or depend on the cloudprovider.

On the other hand, it may not be useful to do the `prom check rules` for each cloudprovider. Maybe we could just do:
- create a list to be checked
- first provider, check all we can and clean them from the list, but if helm extract fails we skip them and keep them in the list to be tried with the next provider.

This process would not care about ignore lists.

Yes, I'll try to do it!

hervenicol (approved):
We discussed it a lot Mario and me, and I think we have the right logic there.
There will probably some bugs & adjustments in the future, but we have a great basis there IMO.

QuentinBisson:
@TheoBrigitte what do you think about merging this ?

helm/prometheus-rules/templates/alerting-rules/capi.rules.yml
```diff
@@ -61,4 +61,4 @@ spec:
           annotations:
             description: |-
               {{`Cluster {{$labels.exported_namespace}}/{{$labels.cluster_name}} is in a non healthy phase.`}}
-{{- end }}
+{{- end }}
```
TheoBrigitte, line 64: Why removing this EOF?

helm/prometheus-rules/.helmignore
TheoBrigitte, line 3: Can't we just ignore the whole `alerting-rules-test/` folder ?

helm/prometheus-rules/templates/alerting-rules/capo.rules.yml
```diff
@@ -49,4 +49,4 @@ spec:
             description: |-
               {{`OpenStackCluster {{$labels.exported_namespace}}/{{$labels.name}} failure reason is {{$labels.reason}}.`}}
             opsrecipe: remove-errors-from-capi-capo-crs/
-{{- end }}
+{{- end }}
```
TheoBrigitte, line 52: Why removing this EOF?

test/hack/bin/fetch-tools.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# let's have reproducable tests - pin to specific versions
+ARCHITECT_VERSION="6.5.0"
+PROMETHEUS_VERSION="2.36.2"
```
TheoBrigitte, line 6: I would rather use the current Prometheus version we are running

```suggestion
PROMETHEUS_VERSION="2.28.1"
```

test/hack/bin/fetch-tools.sh
```diff
@@ -0,0 +1,106 @@
...
+Darwin*)
+    export PROMETHEUS_SOURCE="https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.darwin-amd64.tar.gz"
+    export HELM_SOURCE="https://get.helm.sh/helm-v${HELM_VERSION}-darwin-amd64.tar.gz"
+    export ARCHITECT_SOURCE="https://github.com/giantswarm/architect/releases/download/v${ARCHITECT_VERSION}/architect-v${ARCHITECT_VERSION}-darwin-amd64.tar.gz"
+    export YQ_SOURCE="https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_darwin_arm64.tar.gz"
+    export YQ_BIN_FILE="yq_darwin_arm64"
+    ;;
+
+*)
+    echo "${OS_BASE} not supported - if needed, please implement by your own."
```
TheoBrigitte, line 31: I would suggest to exit early here.

```suggestion
    echo "${OS_BASE} not supported - if needed, please implement by your own."
    exit 1
```

test/hack/bin/fetch-tools.sh
```diff
@@ -0,0 +1,106 @@
...
+    ;;
+
+esac
```
TheoBrigitte, line 34: How about we simply into this

```suggestion
Linux*)
    os="linux"
    ;;

Darwin*)
    os="darwin"
    ;;

*)
    echo "${OS_BASE} not supported - if needed, please implement by your own."
    exit 1
    ;;

esac

export PROMETHEUS_SOURCE="https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.${OS}-amd64.tar.gz"
    export HELM_SOURCE="https://get.helm.sh/helm-v${HELM_VERSION}-${OS}-amd64.tar.gz"
    export ARCHITECT_SOURCE="https://github.com/giantswarm/architect/releases/download/v${ARCHITECT_VERSION}/architect-v${ARCHITECT_VERSION}-${OS}-amd64.tar.gz"
    export YQ_SOURCE="https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_${OS}_amd64.tar.gz"
    export YQ_BIN_FILE="yq_${OS}_amd64"
```

test/hack/bin/verify-rules.sh
```diff
@@ -0,0 +1,200 @@
...
+                promtool_test_output="$("$GIT_WORKDIR/$PROMTOOL" test rules "$provider_testfile" 2>&1)" ||
+                    promtool_test_errors+=("$promtool_test_output")
+            fi
+
+        done
+    done
+
+    # Cleanup
+    git checkout -- "$GIT_WORKDIR/helm/prometheus-rules/Chart.yaml"
+    git checkout -- "$GIT_WORKDIR/helm/prometheus-rules/values.yaml"
```
TheoBrigitte, line 166: I am not really keen on changing the current git state here. I would prefer that all changes happen in a different folder.

test/hack/bin/verify-rules.sh
```diff
@@ -0,0 +1,200 @@
...
+    expected_failure_relative_file_global="test/conf/promtool_ignore"
+    expected_failure_file_global="$GIT_WORKDIR/$expected_failure_relative_file_global"
+
+    # Retrieve all files we're going to check
+    local -a all_files
+    mapfile -t all_files < <(
+        cd "$GIT_WORKDIR" || return 1
+        # filter alerting-rules files, and remove prefix `helm/prometheus-rules/`
+        git ls-files |
+            sed -En 's_^helm/prometheus-rules/(templates/alerting-rules/.*\.ya?ml)$_\1_p' || echo error
```
TheoBrigitte, line 47: That's a bit hidden, and why are we ignoring alerting rules here ?

TheoBrigitte (commented):
I very much like the direction of adding unit test for prometheus rules here. Very good job.

That's a lot of bash foo, I am not super confident we can maintain this. But that's a good start and we easily change this in the future if needed.

test/hack/bin/verify-rules.sh
bavarianbidi: we have to run `architect helm template` to get a valid helm template which is required to generate templated alerting rules.

test/hack/bin/verify-rules.sh
bavarianbidi: `sed -En 's_^helm/prometheus-rules/(templates/alerting-rules/.*\.ya?ml)$_\1_p'` search for all alerting rules (so we exclude all other files expect the alerting rules) and removes the `helm/prometheus-rules` prefix. This is used to run `helm template` for only this file with different providers.

helm/prometheus-rules/templates/alerting-rules/capo.rules.yml
bavarianbidi: happens due rebasing -- reverted the removal

helm/prometheus-rules/templates/alerting-rules/capi.rules.yml
bavarianbidi: happens due rebasing -- reverted the removal

test/hack/bin/verify-rules.sh
TheoBrigitte: I would prefer that we copy the chart folder somewhere in a tmp directory and run `architect helm template` there. Then we can safely trash the test folder.

test/hack/bin/verify-rules.sh
bavarianbidi: i aggree with you, that we should also check the `recording-rules` directory - but i would recommend to implement this after this PR got merged. The testing stuff was primary implemented for alerting-rules and i'm not 100% sure yet how to unit test recording rules.

test/hack/bin/verify-rules.sh
bavarianbidi: just learned that `architect` has the `--working-directory` flag ... should be a easy change, will do that.

test/hack/bin/verify-rules.sh
TheoBrigitte: recording rules are the same things than alerting rules, same format, same resources, just different purpose. But fine let's only test alerting rules for now. This is already a huge improvement.

test/hack/bin/fetch-tools.sh
hervenicol: Would not work as-is because some packages are arm64 for Darwin.
May be a problem though; we should maybe rely on the actual machine's architecture.

hervenicol:
Also, I think before we merge we need:
* a communication because people may complain when they add new rules and it fails because they now need to write tests.
* more tests, so people have examples.

hervenicol:
> That's a lot of bash foo, I am not super confident we can maintain this.

I agree the script is not very easy to read. Some refactoring / functions splitting could be useful for easier maintenance.
But on the other hand, I'd like these tests to be battle-tested ASAP so we have user feedback.

If you have ideas on how to reorganize the code, please tell!

QuentinBisson:
How about we pair on it after the onsite so I can learn bash :)

Le ven. 26 août 2022, 18:25, Hervé Nicol ***@***.***> a
écrit :

hervenicol:
Also, found a bug: the `global` tests don't work, because this is not where the yaml rules files are generated.
Reminder / note to self: global tests have to be run against each provider. So the fix should consist in being in the directory where rules have been generated.

test/hack/bin/verify-rules.sh
TheoBrigitte: I documented in the readme that we only test alerting-rules.

test/hack/bin/fetch-tools.sh
TheoBrigitte: Somehow solved, or I do not care enough.

test/hack/bin/verify-rules.sh
TheoBrigitte: We tried the `--working-directory` flag, which is not working as we expected. It yield a `not a git repository` error. So we decided to keep changing things in the current directory and move the destructive commands under `make clean`.

TheoBrigitte (approved):
We had an extensive session with @hervenicol where we tested the whole thing and improved it.

PR #400, opened 2022-07-20
Improve `CoreDNSDeploymentNotSatisfied` query.

whites11:
in recent releases Coredns is deployed in 2 different deployments, but the old alerting rule was only considering one of the two, leading to false alerts.

This PR attempts at fixing this issue.

### Checklist


njuettner (approved):

whites11:
> Nice! Worth mentioning in the changelog that this change now covers both worker/controlplane deployments?

done, thanks

PR #403, opened 2022-07-20
Adapt rules and alerting to gcp

fiunchinho:
This PR:

- Inhibit gcp alerts out of business hours.
- Don't monitor bastions on gcp.
- Don't monitor vault on gcp.
- Don't monitor cluster-service on gcp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Inhibit gcp alerts out of business hours.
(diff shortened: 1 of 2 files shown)


AverageMarcus (commented):
😬  Oooo I don't like this. Didn't we implement a way of identifying CAPI clusters?

fiunchinho:
If I remember correctly the decision was to use the provider, because most providers will have clusters of one type only, ie `gcp`/`openstack` will have only CAPI clusters. It's just `aws` and `azure` that may have both types, in which case we can use the watch filter label.

AverageMarcus (approved):

calvix (approved):

PR #416, opened 2022-07-25
alerting rules Atlas: alerts that did not page now do

hervenicol:
This PR changes Atlas alerts from "severity: notify" to "severity: page".

Related to:
* https://github.com/giantswarm/giantswarm/issues/22923
* https://github.com/giantswarm/giantswarm/issues/22944
* https://github.com/giantswarm/giantswarm/issues/22920

We had a few tickets where we detected problems without alerts.
The alerting rules were existing, but not paging.

According to alertmanager config, it seems `severity: notify` alerts are sent to Slack, but at Atlas we don't look much at it.
On top of that, the recent alerts I created were supposed to page, but I did not understand the severity setting and misused it.

So I propose to set all the Atlas alerts to `severity: page`, but in order to avoid an increase of area oncall alerts during the night, I also set most of them to only page during workhours.

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus-operator.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus-operator.rules.yml
@@ -36,7 +36,7 @@ spec:
       labels:
         area: empowerment
         cancel_if_outside_working_hours: "true"
-        severity: notify
+        severity: page
         team: atlas
         topic: observability
     - alert: PrometheusOperatorWatchErrors
@@ -47,7 +47,7 @@ spec:
       labels:
         area: empowerment
         cancel_if_outside_working_hours: "true"
-        severity: notify
+        severity: page
         team: atlas
         topic: observability
     - alert: PrometheusOperatorSyncFailed
@@ -58,7 +58,7 @@ spec:
       labels:
         area: empowerment
         cancel_if_outside_working_hours: "true"
-        severity: notify
+        severity: page
         team: atlas
         topic: observability
     - alert: PrometheusOperatorReconcileErrors
@@ -69,7 +69,7 @@ spec:
       labels:
         area: empowerment
         cancel_if_outside_working_hours: "true"
-        severity: notify
+        severity: page
         team: atlas
         topic: observability
     - alert: PrometheusOperatorNodeLookupErrors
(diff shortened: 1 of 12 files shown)


hervenicol:
Deployed on Gauss without any issues. At least, syntax is correct :grin:

helm/prometheus-rules/templates/alerting-rules/prometheus-operator.rules.yml
```diff
@@ -36,7 +36,7 @@ spec:
       labels:
         area: empowerment
         cancel_if_outside_working_hours: "true"
-        severity: notify
+        severity: page
```
mcbenjemaa, line 39: I'm not sure regarding those Prometheus-operator alerts. 
they seem just normal warnings to me, 

Having a notification is just enough for this kind of alerts.
 maybe only email notification is enough. 

So, we can lower the priority of those alerts to P4, P5 ?

What do you think?

helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
```diff
@@ -39,8 +39,8 @@ spec:
         cancel_if_cluster_status_deleting: "true"
         cancel_if_kubelet_down: "true"
         cancel_if_cluster_has_no_workers: "true"
-        cancel_if_outside_working_hours: {{ include "workingHoursOnly" . }}
-        severity: notify
+        cancel_if_outside_working_hours: "true"
```
mcbenjemaa, line 42: I think we keep the default value,

helm/prometheus-rules/templates/alerting-rules/prometheus-operator.rules.yml
hervenicol: For Atlas, there is no such thing as "a warning" "an email" or "a low priority alert".
Either it pages, or not, as far as I understood.

So, IMO
* if we keep it this way we can only see it in the prometheus list of alerts, and in alertmanager.
* or we make it page during WH, so we are informed when it goes wrong
* or we remove the check?

helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
hervenicol: You mean, this alert (`CadvisorDown`) should page outside working hours?
Makes sense, I'll update it.

helm/prometheus-rules/templates/alerting-rules/prometheus-operator.rules.yml
mcbenjemaa: Let’s stick with the current rules,
In future we can enhance it in Opsgenie.

mcbenjemaa (approved):

PR #425, opened 2022-07-28
Add Diskspacetoolow alert of 20% and 10% threshold

mcbenjemaa:
Towards: https://github.com/giantswarm/giantswarm/issues/22959

This PR:

- Add Diskspacetoolow alert of 20% and 10% threshold

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/disk.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/disk.management-cluster.rules.yml
@@ -95,3 +95,27 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: DataDiskPersistentVolumeSpaceTooLow10Percent
+      annotations:
+        description: '{{`The free space on the Data Disk for instance: {{ $labels.instance }} and PVC: {{ $labels.persistentvolumeclaim}} was below 10 percent for longer than 1 hour (current value {{ $value | printf "%.2f" }}).`}}'
+        opsrecipe: low-disk-space/#persistent-volume
+      expr: kubelet_volume_stats_available_bytes/kubelet_volume_stats_capacity_bytes < 0.10
+      for: 1h
+      labels:
+        area: empowerment
+        cancel_if_outside_working_hours: "true"
+        severity: page
+        team: atlas
+        topic: observability
+    - alert: DataDiskPersistentVolumeSpaceTooLow20Percent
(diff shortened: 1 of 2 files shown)


hervenicol:
That would be a great place to play with `predict_linear` :nerd_face: 

Some info here:
https://www.robustperception.io/reduce-noise-from-disk-space-alerts/

From my tests (stuff like `predict_linear(kubelet_volume_stats_available_bytes{instance="172.19.5.229:10250"}[12h],4*3600) <=0`), we can predict a few hours, but not a whole week-end.

mcbenjemaa:
The question for now:

- Should we keep the current alert?
- Having both alerts or just one is enough?

mcbenjemaa:
@QuentinBisson Would you take a look?

helm/prometheus-rules/templates/alerting-rules/disk.management-cluster.rules.yml
```diff
@@ -95,3 +95,27 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: DataDiskPersistentVolumeSpaceTooLow10Percent
```
QuentinBisson, line 98: Do we really want to get alerted for all the disks ? And I'm not sure we need both alerts, we should probably keep the 10% one

hervenicol: * Alert for all the data disks: Isn't it what the existing alerts are already supposed to do? https://github.com/giantswarm/giantswarm/issues/22959
* Threshold: I think too we only need 1 alert. But I don't know what is the right threshold. I'd chose the 10% one, but yet we may receive a lot of alerts.

mcbenjemaa: Yeah, I was just thinking most of the apps that use PVs are managed by us: (Prometheus, Alertmanager, EFK ...)

mcbenjemaa: Ok lemme fix it.

mcbenjemaa: can you check

QuentinBisson (approved):

QuentinBisson:
Looking good :)

PR #430, opened 2022-08-03
CadvisorDown

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/23029

This PR add an inhibition when kubelet are down and use this to silence CadvisorDown when needed. Also extend delay to 1h, and make it page only during working hours.

* extend CadvisorDown delay to 1h
* bring back KubeletDown inhibition
* CadvisorDown only page during working hours

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 
-- Inhibit CadvisorDown with InhibitionKubeletDown
-- Extend delay CadvisorDown to 1h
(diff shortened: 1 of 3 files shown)


TheoBrigitte:
Confirmed that cadvisor go down the same time as kubelet goes down.

I manually stopped a kubelet and here is the result :

image
image

And obviously the alert is triggered (shows as pending because 15mn are not yet over)
image

TheoBrigitte:
With this change the `CadvisorDown` alert is inhibited by `InhibitionKubeletDown` as soon as a kubelet is down.

image

I doubt we ever see this alert paging us again, but still good to have.

hervenicol (approved):
LGTM

PR #444, opened 2022-08-17
extending waiting period for AWSClusterUpdateFailed

njuettner:
Larger clusters might take longer to update, depending on the size of
the cluster and the ASG settings of pause time between rolling nodes.

Issue: https://github.com/giantswarm/giantswarm/issues/23291


njuettner:
Going with a ping ✌🏻

PR #452, opened 2022-08-25
cert manager cert requests

mcharriere:
This PR:

Add CertManagerTooManyCertificateRequests alert to identify early possible control plane overloading.

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
@@ -46,7 +46,7 @@ spec:
       expr: sum by (cluster_id) (etcd_kubernetes_resources_count{kind="certificaterequests.cert-manager.io"}) > 10000
       for: 15m
       labels:
-        area: managedapps
+        area: managedservices
         cancel_if_outside_working_hours: "true"
         severity: notify
         team: cabbage
--- a/helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
@@ -39,3 +39,15 @@ spec:
         severity: page
         team: cabbage
         topic: cert-manager
+    - alert: CertManagerTooManyCertificateRequests
+      annotations:
+        description: '{{`There are too many CertificateRequests in cluster {{ $labels.cluster_id }}.`}}'
+        opsrecipe: cert-requests-too-many/
+      expr: sum by (cluster_id) (etcd_kubernetes_resources_count{kind="certificaterequests.cert-manager.io"}) > 5000
+      for: 15m
+      labels:
+        area: managedapps
+        cancel_if_outside_working_hours: "true"
+        severity: notify
+        team: cabbage
+        topic: cert-manager
--- a/helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
@@ -43,7 +43,7 @@ spec:
       annotations:
         description: '{{`There are too many CertificateRequests in cluster {{ $labels.cluster_id }}.`}}'
         opsrecipe: cert-requests-too-many/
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
```diff
@@ -39,3 +39,15 @@ spec:
...
         team: cabbage
         topic: cert-manager
+    - alert: CertManagerTooManyCertificateRequests
+      annotations:
+        description: '{{`There are too many CertificateRequests in cluster {{ $labels.cluster_id }}.`}}'
+        opsrecipe: cert-requests-too-many/
+      expr: sum by (cluster_id) (etcd_kubernetes_resources_count{kind="certificaterequests.cert-manager.io"}) > 5000
+      for: 15m
+      labels:
+        area: managedapps
```
mcharriere, line 49: is managedapps ok?

helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
```diff
@@ -39,3 +39,15 @@ spec:
         severity: page
         team: cabbage
         topic: cert-manager
+    - alert: CertManagerTooManyCertificateRequests
+      annotations:
+        description: '{{`There are too many CertificateRequests in cluster {{ $labels.cluster_id }}.`}}'
+        opsrecipe: cert-requests-too-many/
+      expr: sum by (cluster_id) (etcd_kubernetes_resources_count{kind="certificaterequests.cert-manager.io"}) > 10000
```
mcharriere, line 46: 10000 is a completely arbitrary number. During the previous incident, there were ~75k CertificateRequests. Unfortunately we didn't have metrics to set a better number.

Now we are collecting metrics from most of the WC in viking (ref https://github.com/giantswarm/giantswarm/issues/22495) and we should have a better picture to set a new threshold in the future.

ubergesundheit:
Since we have metric for the count of all objects in etcd, wouldn't it make sense to alert on all resources? Or is it that a single kind of resource causes problem not the whole number? My thinking is that this can happen for other resource types and not just cert-manager

mcharriere:
> Since we have metric for the count of all objects in etcd, wouldn't it make sense to alert on all resources? Or is it that a single kind of resource causes problem not the whole number? My thinking is that this can happen for other resource types and not just cert-manager

Although I don't where the limit on the number of resources is, it's true that it might be a problem. Having said that, I believe that alerting on all the resources has a few drawback and we should not pursue it:
1. It's hard to define an owner, as the scope is too wide.
2. It's impossible to define a recipe for resolution during on-call. `TooManyResources` feels more like a capacity planning issue than a "the cluster is broken, fix it" thing.

Also, I think the alert makes sense for this specific case because the number of `CertificateRequests` is always going up. Once the cleanup mechanism is properly working, this alert has way less importance.

helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
Gacko: `nginx-ingress-controller` and `external-dns` - for which Cabbage is also responsible for - uses `managedservices`.

Gacko: https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/alerting-rules/external-dns.rules.yml
https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/alerting-rules/ingress-controller.rules.yml

Gacko (changes requested):

helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
mcharriere:
```suggestion
        area: managedservices
```

helm/prometheus-rules/templates/alerting-rules/cert-manager.rules.yml
TheoBrigitte: That's fine. We have a lot of "arbitrary" number in our alerting that comes from experience.

TheoBrigitte (approved):

Gacko (approved):

PR #455, opened 2022-08-30
include namespace in alert description

kubasobon:
This PR:

- Includes namespace in `WorkloadClusterAppFailed` alert's description

### Checklist

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
@@ -143,7 +143,7 @@ spec:
     ## By default they go to Honeybadger, unless one of the teams below is configured.
     - alert: WorkloadClusterAppFailed
       annotations:
-        description: '{{`Workload Cluster App {{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
+        description: '{{`Workload Cluster App {{ $labels.exported_namespace }}/{{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
         opsrecipe: app-failed/
       expr: label_replace(app_operator_app_info{status!~"(?i:(deployed|cordoned|not-installed))", catalog!~"control-plane-.*", team!~"(?i:(atlas|cabbage|phoenix|hydra|rocket))"}, "cluster_id", "$1", "namespace", "(([^g]|g[^i]|gi[^a]|gia[^n]|gian[^t]|giant[^s]|giants[^w]|giantsw[^a]|giantswa[^r]|giantswar[^m])*)") == 1
       for: 30m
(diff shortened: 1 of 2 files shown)


uvegla (approved):

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -143,7 +143,7 @@ spec:
     ## By default they go to Honeybadger, unless one of the teams below is configured.
     - alert: WorkloadClusterAppFailed
       annotations:
-        description: '{{`Workload Cluster App {{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
+        description: '{{`Workload Cluster App {{ $labels.exported_namespace }}/{{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
```
hervenicol, line 146: I guess it's just `namespace` rather than `exported_namespace`.
At least from the labels I've had on `gauss` with this request.

kubasobon: `$labels.namespace` is often overridden. This was the case for me in many alerts. Let's take [this alert](https://giantswarm.app.opsgenie.com/alert/detail/18c3ee0c-2a08-4603-be2f-701bd26a8bba-1661868811823/details) as an example. The `namespace` is `giantswarm`, but `exported_namespace` is `dmec3`, which is the cluster's name and the namespace in which the failing app can be found. The same applies to CAPX clusters, where `exported_namespace` equals organization namespace. That's the proper namespace where the App CR is.

QuentinBisson: We decided in atlas to remove the exported_namespace so we need to fix the source of the metric

QuentinBisson: So the issue is the app exporter already has a namespace label but a new one is added via relabelling which sets the existing namespace label to exported_namespace. We need to fix this

QuentinBisson: I'll try this https://github.com/giantswarm/kyverno-policies-observability/pull/27

kubasobon: Consider this PR on hold, then.

QuentinBisson: PR works but kyverno is configured to ignore it's own namespace which is the giantswarm namespace and is excluding a lot of service monitors. As a suggestion until this is resolved, can we change the labels to:

```{{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}```

kubasobon: Good suggestion, thanks! I'll update my PR.

QuentinBisson (approved):

QuentinBisson:
Thanks @kubasobon :)

PR #460, opened 2022-09-07
Lowered ManagedLoggingElasticsearchDataNodesNotSatisfied alert sensitivity

hervenicol:
This PR:
- Lowers `ManagedLoggingElasticsearchDataNodesNotSatisfied` alert sensitivity - now triggers after 15min to avoid triggering on pod restarts
- 
### Checklist


webwurst (approved):
Question here is if the customer gets alerted before us and pings in our channels. But if the current setting is an issue for your team, probably fine to try out.

PR #465, opened 2022-09-12
Add PrometheusJobFailure alerting rule

marieroque:
Towards https://github.com/giantswarm/giantswarm/issues/23435

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
@@ -52,3 +52,16 @@ spec:
         team: atlas
         topic: observability
         cancel_if_outside_working_hours: "true"
+    - alert: PrometheusJobFailures
+      annotations:
+        description: {{`Prometheus {{$labels.installation}}/{{$labels.cluster_id}} has failed to scrape all targets in {{$labels.job}} job.`}}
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
```diff
@@ -52,3 +52,16 @@ spec:
         team: atlas
         topic: observability
         cancel_if_outside_working_hours: "true"
+    - alert: PrometheusJobFailures
```
QuentinBisson, line 55:
```suggestion
    - alert: PrometheusJobScrapingFailure
```

Is that name not more meaningful?

TheoBrigitte: Maybe, but I also dislike longer names. Just a preference.

QuentinBisson: Why do you dislike longer names? Being able to know at a glance what this alert is makes everything easier

QuentinBisson: As an example: 
PrometheusJobFailures hints that a job (i.e. cronjob) named Prometheus fails to run or is failing, having the notion of scraping ensure we knowwhat this is about :)

TheoBrigitte (approved):

PR #472, opened 2022-09-23
Make WC pod restart alert filter by namespace

pipo02mix:
This PR:

- adds/changes/removes etc

### Checklist


pipo02mix:
I have found we were receiving alerts for customer workloads. I think we should only alert if our managed containers are restarted

glitchcrab (approved):
lgtm

glitchcrab:
ref https://github.com/giantswarm/giantswarm/issues/23779

PR #486, opened 2022-10-10
Change rules, so they route according to app ownership reported by app-exporter

piontec:
- App CR related (coming from app-exporter) alerting rules are changed:
  - Per-team duplicated rules are removed and only one rule per alert type is kept. It doesn't overwrite the
    incoming `team` label, so alerts are always routed according to the `team` label value coming from `app-exporter`
- unassigned team is now detect using empty or 'noteam' value of the `team` label

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
@@ -137,7 +137,7 @@ spec:
         cancel_if_cluster_status_updating: "true"
         cancel_if_outside_working_hours: "true"
         severity: page
-        team: honeybadger
+        # team: honeybadger <-- don't overwrite the team, leave as it was, so it will go to the already labeled team
         topic: releng
     ## Workload cluster app rules only available on the management cluster Prometheus because the source is app-exporter.
     ## By default they go to Honeybadger, unless one of the teams below is configured.
--- a/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
@@ -25,7 +25,7 @@ spec:
         cancel_if_outside_working_hours: "true"
         severity: page
         sig: none
-        team: atlas
+        team: atlas   # <-- why we don't just leave the team label and stop the per=team alerting rules?
         topic: releng
     - alert: ManagementClusterAppFailedCabbage
       annotations:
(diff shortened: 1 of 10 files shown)


kubasobon (approved):
LGTM

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -137,115 +40,30 @@ spec:
...
-        sig: none
-        team: phoenix
-        topic: releng
-    - alert: WorkloadClusterAppFailedHydra
-      annotations:
-        description: '{{`Workload Cluster App {{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
-        opsrecipe: app-failed/
-      expr: label_replace(app_operator_app_info{status!~"(?i:(deployed|cordoned|not-installed))", catalog!~"control-plane-.*", team="hydra"}, "cluster_id", "$1", "namespace", "(([^g]|g[^i]|gi[^a]|gia[^n]|gian[^t]|giant[^s]|giants[^w]|giantsw[^a]|giantswa[^r]|giantswar[^m])*)") == 1
+      # we have this Frankenstein for namespace checks because go (and prometheus) don't support negative lookaheads in regex
+      expr: label_replace(app_operator_app_info{status!~"(?i:(deployed|cordoned|not-installed))", catalog="giantswarm", team!~"^$|noteam"}, "cluster_id", "$1", "namespace", "(([^g]|g[^i]|gi[^a]|gia[^n]|gian[^t]|giant[^s]|giants[^w]|giantsw[^a]|giantswa[^r]|giantswar[^m])*)") == 1
```
QuentinBisson, line 49: Do you mind moving the frankenstein expression to helpers to avoid mistypes? Otherwise, I'm fine with it. Can someone else from @giantswarm/team-atlas take a look?

piontec: Good idea :)

ljakimczuk (approved):

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
hervenicol: Would be great to have a unit test testing this replacement expression.

PR #494, opened 2022-10-23
Add SLO alert for Cilium Agent.

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/1510

This PR:

- adds SLO alerting for cilium agent on all management clusters

this is an SLO alert, meaning it can't be made to page only during business hours.
The threshold is arbitrary and it might be too noisy, but there is no other way but test this.

### Checklist

- [ ] Add Unit tests


fiunchinho (approved):
We can always tweak it later on

PR #533, opened 2022-11-30
change prometheus agent alert to be on MC as well

mcbenjemaa:


QuentinBisson:
We need to change the inhition alert then because this will page for all vintage MCs

QuentinBisson:
We can probably check if the provider is aws, azure, kvm and if it is we inhibit for the mc

QuentinBisson (approved):

PR #540, opened 2022-12-01
Improve prometheus-agent inhibition

mcbenjemaa:
towards https://github.com/giantswarm/giantswarm/issues/24807

### Summary

Change the inhibition to 🔥  on each cluster, 
Even if MC is down the inhibition is still firing for WC and thus will inhibit alerts of WCs...

#### Explanation 

We compare the list of `prometheus_build_info{app="prometheus"}` with the list of `prometheus_build_info{app="prometheus",instance="prometheus-agent"}`

we get a clusterID, If a cluster missing `instance="prometheus-agent"`.

### Testing

TBD

/cc @hervenicol

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/inhibit.prometheus-agent.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/inhibit.prometheus-agent.rules.yml
@@ -18,7 +18,7 @@ spec:
     - alert: InhibitionClusterIsNotRunningPrometheusAgent
       annotations:
         description: '{{`Cluster ({{ $labels.cluster_id }}) is not running Prometheus Agent.`}}'
-      expr: count by (cluster_id) (prometheus_build_info{app="prometheus"}) unless count by (cluster_id) (prometheus_build_info{app="prometheus",instance="prometheus-agent"})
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/alerting-rules/inhibit.prometheus-agent.rules.yml
```diff
@@ -11,15 +11,14 @@ spec:
   - name: inhibit.prometheus-agent
     rules:
     # this inhibition fires when a cluster is not running prometheus-agent.
-    # we retrieve the list of existing cluster IDs from `kube_namespace_created`
-    # then compare it with the list of deployed observability-bundles from `app_operator_app_info`
+    # we retrieve the list of existing cluster IDs from `prometheus_build_info` with app="prometheus"
+    # then compare it with the list of `prometheus_build_info` with instance="prometheus-agent"
```
hervenicol, line 15: I think now the syntax has been simplified, we could also simplify the explanation:
```suggestion
    # If we we find prometheus_build_info for instance="prometheus-agent", it means prometheus-agent is running on this cluster
    # so, raise an inhibition unless prometheus-agent runs on the cluster 
```

but, do we want to inhibit when `prometheus-agent` runs on the cluster (current inhibition) or when `prometheus-agent` **is supposed to** run on the cluster (previous version)?

mcbenjemaa: we just want to inhibit if the cluster is not running prometheus-agent

hervenicol (approved):

PR #553, opened 2022-12-08
unit tests: add aws and fix tests

hervenicol:
This PR adds unit tests for AWS.

By doing so, I discovered that `workingHoursOnly` template was a hard "if it's GCP or Openstack" condition.

This means:
* I don't see the point in it
* it makes tests behave different in unexpected ways on different providers
* it makes alerting behave different in unexpected ways on different providers

So, for tests to be consistent, I removed it from the rules that already have tests running, and set the "default value for most providers" (ie `false`) instead.

### Checklist

- [x] Add Unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- unit tests for AWS
+
+### Changed
+
+- removed "workingHoursOnly" template from loki and crossplane tests
+
 ## [2.67.0] - 2022-12-08
 
 ### Changed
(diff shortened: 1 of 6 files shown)


uvegla:
For crossplane we definitely do not want alerts to outside business hours as it is in a highly experimental stage at the moment. Is there an alternative to disable the alerts for night shifts?

hervenicol:
> For crossplane we definitely do not want alerts to outside business hours as it is in a highly experimental stage at the moment. Is there an alternative to disable the alerts for night shifts? Or just let's simply set it to `true` for crossplane, please.

OK, done.

For the record: it was not yet paging because only deployed on testing installations AFAIK.
But the alert, the way it was defined before, was set to "page on stable installations except GCP or Openstack".
The whole `workingHoursOnly` subject is explained here: https://gigantic.slack.com/archives/C01F7T2MNRL/p1670573463468649 ... and we may clean it up when we have an answer from Hydra.

uvegla (approved):

PR #554, opened 2022-12-12
Fix kubelet down inhibition

QuentinBisson:
This PR:

- fixes kubelet down inhibition and it's application towards https://github.com/giantswarm/giantswarm/issues/24746

### Checklist

- [ ] Add Unit tests. sadly not today

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/inhibit.all.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/inhibit.all.rules.yml
@@ -21,7 +21,7 @@ spec:
         team: phoenix
         topic: monitoring
     - alert: InhibitionKubeletDown
-      expr: up{app="kubelet"} == 0
+      expr: label_replace(up{app="kubelet"}, "ip", "$1", "instance", "(.+):\\d+") == 0
       labels:
         kubelet_down: true
         area: kaas
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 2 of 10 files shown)


helm/prometheus-rules/templates/alerting-rules/inhibit.all.rules.yml
```diff
@@ -21,7 +21,7 @@ spec:
         team: phoenix
         topic: monitoring
     - alert: InhibitionKubeletDown
-      expr: up{app="kubelet"} == 0
+      expr: label_replace(up{app="kubelet"}, "ip", "$1", "instance", "(.+):\\d+") == 0
```
hervenicol, line 24: Turns out `up{app="cadvisor"}` has the `node` value, and the same `instance` value.
Wouldn't it be nice to have all the heavy lifting in the inhibition, so other rules remain simpler to maintain?

I think this would do the job:
 ```suggestion
            label_replace(up{app="kubelet"} * on(instance) group_left(node) up{app="cadvisor"}, "ip", "$1", "instance", "(.+):\\d+")
```

QuentinBisson: This means the inhibition would be on cluster_id, node instead of cluster_id, ip right?

hervenicol: oh, yes, my bad, we can't easily match on either one or the other. forget about my comment :(

hervenicol (approved):
:+1:

helm/prometheus-rules/templates/alerting-rules/inhibit.all.rules.yml
QuentinBisson: We really need to discuss inhibitions. I added an issue draft on the board but let's discuss tomorrow

PR #566, opened 2022-12-15
Fix boolean in rules

QuentinBisson:
This PR:

- fixes booleans

#### Redeploying app prometheus-rules ####
Error: UPGRADE FAILED: failed to create resource: PrometheusRule.monitoring.coreos.com "crossplane.rules" is invalid: [spec.groups[0].rules[0].labels.cancel_if_outside_working_hours: Invalid value: "boolean": spec.groups[0].rules[0].labels.cancel_if_outside_working_hours in body must be of type string: "boolean", spec.groups[0].rules[1].labels.cancel_if_outside_working_hours: Invalid value: "boolean": spec.groups[0].rules[1].labels.cancel_if_outside_working_hours in body must be of type string: "boolean"]

### Checklist


hervenicol (changes requested):
Booleans should be booleans.
I don't see no reason why helm or prometheus would deny them.

Looks like it works on most installations.
Works on our unit tests (that use helm and prometheus).

So before we do this change, I'd like to understand why it breaks, and if it has a valid reason for that.

marieroque:
How can i reproduce the issue ?

QuentinBisson:
```sh
#!/usr/bin/env bash

echo start

#for i in $(opsctl list installations -s); do
#	echo $i
#done
#exit

#set -x

dryRun=${1:-true}
if [[ $dryRun != false ]]; then
	dryRun=false
fi

get_info_app() {
    app="$1"
    context="$2"
    
    extraArgs=""
    if [[ $context ]]; then
      extraArgs="--context=gs-$context"
    fi

    chart="$(kubectl get chart -n=giantswarm "$app" "$extraArgs" -oyaml | yq -o j)" || return 1
    [[ "$chart" == "null" ]] && return 1
    url="$(echo "$chart" | jq -r '.spec.tarballURL')"
    appNamespace="$(echo "$chart" | jq -r '.spec.namespace')"
}

recreate_app() {
    app="$1"
    context="$2"
    get_info_app "$app" "$context" || return

    extraArgs=""
    if [[ $context ]]; then
      extraArgs="--kube-context=gs-$context"
    fi

    # set -x
    (
      echo ""
      echo "#### Redeploying app $app ####"
      helm upgrade -n "$appNamespace" "$app" "$url" --reuse-values --force "$extraArgs"
    )
}

list_apps() {
    label="$1"
    context="$2"
    extraArgs=""
    if [[ $context ]]; then
      extraArgs="--context=gs-$context"
    fi

    charts=$(kubectl get chart -l "$label" $extraArgs -n giantswarm -o custom-columns=NAME:metadata.name --no-headers)
    echo $charts
}

redeploy_promop_apps_mc() {
    context="$1"
    appOperatorApps=$(list_apps app.kubernetes.io/name=app-operator $context)
    apps=(prometheus-rules prometheus-meta-operator $appOperatorApps app-exporter dex-app grafana kyverno kyverno-policies kyverno-policies-observability kyverno-policies-ux prometheus-operator-app prometheus-agent macropower-analytics-panel-server starboard-exporter)
    #for app in "${apps[@]}"; do
        # recreate_app "$app" "$context"
    #done
    recreate_app prometheus-rules "$context"
}


main() {
    context="$1"

    #recreate_app grafana
    #recreate_app kyverno-policies-observability
    redeploy_promop_apps_mc $context
    # recreate_app kyverno
}

main "$@"
```

You can run this script on some installations using ./script.sh installation-name. You need to run opsctl login installation-name first

hervenicol:
Oh, these are labels.
And from the CRD (https://github.com/giantswarm/prometheus-operator-crd/blob/master/helm/prometheus-operator-crd/templates/crd-prometheusrules.yaml#L80) all labels should be strings.

I still don't understand why it fails on some clusters and not on some others. But for sure as these are labels and annotations, OK to force them as strings.

Then, this fixes current ones, but nothing prevents us from putting booleans again in the future.

QuentinBisson:
The thing is true can be interpreted as a string and a boolean. Could it be something different happening with webhooks on Gauss?

hervenicol (approved):
I changed my mind :grin:

hervenicol:
> The thing is true can be interpreted as a string and a boolean. Could it be something different happening with webhooks on Gauss?

"Labels should be strings" was enough to convince me.
But for the sake of science, I'd love to understand what's happening.
Is there a way to disable webhooks so we can check this?

QuentinBisson:
You can remove them :D

QuentinBisson:
using kubectl get -oyaml (validating and mutating)  in a file then remove them from the cluster and apply them back once you're done :)

QuentinBisson:
Feel free to merge this one when you're done with your tests :)

hervenicol:
It's not becaue of a cpol :shrug:

PR #578, opened 2022-12-28
Fix KubeStateMetricsDown alert

mcbenjemaa:
Towards https://github.com/giantswarm/giantswarm/issues/22262

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
@@ -48,33 +48,15 @@ spec:
       annotations:
         description: '{{`KubeStateMetrics ({{ $labels.instance }}) is down.`}}'
         opsrecipe: kube-state-metrics-down/
-      expr: label_replace(up{app="kube-state-metrics"}, "ip", "$1.$2.$3.$4", "node", "ip-(\\d+)-(\\d+)-(\\d+)-(\\d+).*") == 0
+      expr: label_replace(up{app="kube-state-metrics"}, "ip", "$1.$2.$3.$4", "node", "ip-(\\d+)-(\\d+)-(\\d+)-(\\d+).*") == 0  or absent(up{app="kube-state-metrics"} == 1)
       for: 15m
       labels:
         area: kaas
         cancel_if_apiserver_down: "true"
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
-        cancel_if_any_kube_state_metrics_down: "true"
         cancel_if_cluster_has_no_workers: "true"
         cancel_if_kubelet_down: "true"
-        cancel_if_outside_working_hours: {{ include "workingHoursOnly" . }}
-        severity: page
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
```diff
@@ -48,34 +48,16 @@ spec:
       annotations:
         description: '{{`KubeStateMetrics ({{ $labels.instance }}) is down.`}}'
         opsrecipe: kube-state-metrics-down/
-      expr: label_replace(up{app="kube-state-metrics"}, "ip", "$1.$2.$3.$4", "node", "ip-(\\d+)-(\\d+)-(\\d+)-(\\d+).*") == 0
+      expr: label_replace(up{app="kube-state-metrics"}, "ip", "$1.$2.$3.$4", "node", "ip-(\\d+)-(\\d+)-(\\d+)-(\\d+).*") == 0  or absent(up{app="kube-state-metrics"} == 1)
```
TheoBrigitte, line 51: What's this magic with the node ip ? Do we need to propagate this ?

mcbenjemaa: I'm not sure.. .

mcbenjemaa: all up alerts are the same

mcbenjemaa: Do we merge?

TheoBrigitte: I think this label might be used to match against other metrics, so I believe we should keep it.

hervenicol: I did not check, but I'd bet on matching inhibitions.

mcbenjemaa: @QuentinBisson  has more knowledge about this

hervenicol: I confirm it is inhibitions:

```
- source_matchers:
  - kubelet_down="true"
  target_matchers:
  - cancel_if_kubelet_down="true"
  equal:
  - cluster_id
  - ip
```
and:
```
- source_matchers:
  - kubelet_not_ready="true"
  target_matchers:
  - cancel_if_kubelet_not_ready="true"
  equal:
  - cluster_id
  - ip
```

hervenicol:
I think we perverted the way inhibitions should work.

We have a 2 checks that looks for `kube-state-metrics` status:
* One called `InhibitionKubeStateMetricsDown`.
* One called `KubeStateMetricsDown`.

:shrug: That's the way all of our inhibitions work. I guess it's supposed to make it easier to read.
But I don't see why "inhibition" checks should not page.

hervenicol (approved):
LGTM.
I have concerns, but on a broader level.
No reasons to mix them with this PR :+1:

PR #584, opened 2023-01-02
prometheus-agent rule: add dashboard link (prometheus remote-write)

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/25049

This is the dashboard I used the most when debugging prometheus-agent, thought it would be worth adding it to the alert.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus-agent.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus-agent.rules.yml
@@ -17,6 +17,7 @@ spec:
         description: '{{`Prometheus agent remote write is failing.`}}'
         summary: Prometheus agent fails to send samples to remote write endpoint.
         opsrecipe: prometheus-agent-remote-write-failed/
+        dashboard: BG-1wO5Vk/prometheus-remote-write
       expr: absent_over_time(up{instance="prometheus-agent"}[10m])
       for: 10m
       labels:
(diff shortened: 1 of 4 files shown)


helm/prometheus-rules/templates/alerting-rules/prometheus-agent.rules.yml
```diff
@@ -17,6 +17,7 @@ spec:
         description: '{{`Prometheus agent remote write is failing.`}}'
         summary: Prometheus agent fails to send samples to remote write endpoint.
         opsrecipe: prometheus-agent-remote-write-failed/
+        dashboard: BG-1wO5Vk/prometheus-remote-write
```
marieroque, line 20: Is the URL the same between each MC ?

hervenicol: The dashboard UID and the dashboard name are the same, yes.
Alertmanager will build the address pointing to the installation's grafana.

hervenicol: Argh, you're right:
Most dashboards have a `uid` defined (see https://github.com/giantswarm/dashboards/blob/master/helm/dashboards/dashboards/shared/private/credentials.json#L419), but some don't.
For instance, prometheus-remote-write does not have it defined.
So, grafana generates a random one.

So, I'll have to:
- update dashboard to have a `uid`
- update `prometheus-rules` documentation with this info

QuentinBisson (approved):

QuentinBisson:
So, do we need dashboard uid guidelines?

hervenicol:
> So, do we need dashboard uid guidelines?

Good question.
Today most dashboards have one because we create them in grafana, then export them. So they keep the UID grafana gave them.
But for generated dashboards (mixins) that's probably gonna be different, we may have to add them another way.

I'm not sure that's an issue for the moment, I think we can deal with this later.

PR #586, opened 2023-01-03
Add Harbor rules

William-Young-97:
Added harbor.rules.yml file with 3 alerts (InstanceDown, HarborUnhealthy and  ProjectQuotaIsRaisingTheLimit).


hervenicol:
Build fails because there were no unit tests written for this new rules.
See https://github.com/giantswarm/prometheus-rules#testing for how to write them.

Hint: to go faster when running the tests, you can do `make test test_filter=harbor`

helm/prometheus-rules/templates/alerting-rules/harbor.all.rules.yml
```diff
@@ -0,0 +1,43 @@
...
+        severity: critical
+        app: harbor
+    - alert: ProjectQuotaIsRaisingTheLimit
+      annotations:
+        summary: The quota of the repository '{{` {{ $labels.repo_name }} is getting close to the limit. `}}'
+      expr: harbor_quotas_size_bytes{type="used"} / ignoring(type) harbor_quotas_size_bytes{type="hard"} > 0.95
+      for: 5m
+      labels:
+        severity: critical
+        app: harbor
```
MarcelMue, line 43: Newline please

test/tests/providers/global/harbor.all.rules.test.yml
```diff
@@ -0,0 +1,24 @@
+---
```
MarcelMue, line 1: Not needed I think

test/tests/providers/global/harbor.all.rules.test.yml
```diff
@@ -0,0 +1,24 @@
...
+              app: harbor
+              cancel_if_apiserver_down: "true"
+              cancel_if_cluster_status_creating: "true"
+              cancel_if_cluster_status_deleting: "true"
+              cancel_if_cluster_status_updating: "true"
+              cancel_if_scrape_timeout: "true"
+              cancel_if_outside_working_hours: "true"
+            exp_annotations:
+              description: " of job  has been down for more than 5 minutes."
+              summary: " Instance  down."
```
MarcelMue, line 24: Newline

helm/prometheus-rules/templates/alerting-rules/harbor.all.rules.yml
```diff
@@ -0,0 +1,43 @@
...
+        cancel_if_outside_working_hours: "true"
+    - alert: HarborUnhealthy
+      annotations:
+        description: '{{` {{ $labels.instance }} is unhealthy.`}}'
+        summary: "Habor cluster is unhealthy."
+      expr: harbor_health == 0
+      for: 5m
+      labels:
+        severity: critical
+        app: harbor
```
MarcelMue, line 35: Should these also cancel on the conditions above?

test/tests/providers/global/harbor.all.rules.test.yml
```diff
@@ -0,0 +1,24 @@
...
+rule_files:
+  - harbor.all.rules.yml
+tests:
+  - interval: 1m
+    input_series:
+      - series: 'harbor_up{}'
+        values: "_x5 1+0x5 0+0x5"
+    alert_rule_test:
+      - alertname: HarborDown
+        eval_time: 5m
```
hervenicol, line 10: Your values say:
* from t0 to t4, no data
* from t5 to t9, "harbor_up=1"
* from t10 to t14 "harbor_down=0"

At `t5`, do you really expect an alert to fire?
I'm not sure data is added before or after the test, but either you have no data (ie probably harbor is not installed) or you have just one data point saying it's running.

This means, on clusters where harbor is not installed this alert will fire?

William-Young-97:
@MarcelMue Should be good to go now pending review.

test/tests/providers/global/harbor.all.rules.test.yml
William-Young-97: Would it be a good idea to have these unit tests as separate series and values?

William-Young-97:
@hervenicol Something like this?

hervenicol:
> @hervenicol Something like this?

Yes, looking good for `HarborDown` tests! :+1: 
Could you apply the same logic to other alerts tests as well?

hervenicol:
The PR looks good from a technical point of view.

Now it's missing a few things as per https://github.com/giantswarm/prometheus-rules/#how-alerts-are-structured:

- [ ] `CHANGELOG.md` update
- [ ] `opsrecipe` annotation
- [ ] `dashboard` annotation (if you have some Grafana dashboard for Harbor)
- [ ] `area` label
- [ ] `team` label

...but it may be more @MarcelMue 's job for some.

MarcelMue:
Needs to be finalized by @giantswarm/team-honeybadger before merge.

QuentinBisson:
@MarcelMue do we still need to keep this PR?

PR #599, opened 2023-01-12
check ca errors only for cloud provider errors

njuettner:
Related issue: https://github.com/giantswarm/giantswarm/issues/25286

Proposal to make this alert less noisy:

We only include Cloud Provider Errors (my most favourite ones), which only checks for things like:

- Failing to get node group info
- Failing to set node group size
- ...

And we exclude errors

- Internal errors
- API call errors (K8s API Server)

Those should be covered by other alerts.


whites11:
> Related issue: [giantswarm/roadmap#1855](https://github.com/giantswarm/roadmap/issues/1855)
> 
> Proposal to make this alert less noisy:
> 
> We only include Cloud Provider Errors (my most favourite ones), which only checks for things like:
> 
> * Failing to get node group info
> * Failing to set node group size
> * ...
> 
> And we exclude errors
> 
> * Internal errors
> * API call errors (K8s API Server)
> 
> Those should be covered by other alerts.

not sure. I still think we should alert on the outcome, not the process. I'm ok with any number of errors connecting to the cloud provider's API as long as the cluster is not needing any more/less node. See what I mean?

njuettner:
As discussed I added 2 more alerts

`cluster_autoscaler_unneeded_nodes_count` - checks if cluster is able to scale in, if not there's something fishy and we need to check
`cluster_autoscaler_unschedulable_pods_count` - checks if pods can be scheduled or stuck in "Pending"

We need to check if this get's too noisy and if the checks are helpful.

helm/prometheus-rules/templates/alerting-rules/cluster-autoscaler.rules.yml
```diff
@@ -16,7 +16,7 @@ spec:
       annotations:
         description: '{{`Cluster-Autoscaler on {{ $labels.cluster_id }} has increased errors.`}}'
         opsrecipe: cluster-autoscaler-scaling/
-      expr: increase(cluster_autoscaler_errors_total[10m]) > 0
+      expr: increase(cluster_autoscaler_errors_total{type="cloudProviderError"}[10m]) > 0
```
whites11, line 19: since we added the two new alerts, I would ditch this one.
as long as pods are scheduled and there are no unneeded nodes we don't care if cluster autoscaler is erroring.

wdyt?

njuettner: Sounds good let me revert this

whites11 (approved):
let's go

PR #608, opened 2023-01-17
Update promtool and fix makefile

mcbenjemaa:
Now we're running Prometheus 2.41 on our clusters, we can upgrade promtool for the tests too.


hervenicol (approved):
LGTM.
Please make sure prometheus 2.41 upgrade is finished before merging.
Otherwise it may not catch errors when we use a too recent promql syntax (like some new relabelling features that were added to 2.41).

hervenicol:
@QuentinBisson  :arrow_up:

QuentinBisson:
PMO was released yesterday so it's fine. The agent is on 2.40.5 but rules are not applying to it

PR #643, opened 2023-02-07
Add container memory/cpu usage for MCs to cortex

teemow:
This adds aggregated memory and CPU usage of apps on the management cluster to cortex. 

Towards: https://github.com/giantswarm/giantswarm/issues/25709

### Checklist

- [x] Update CHANGELOG.md


hervenicol (approved):
LGTM.
However, I'm not sure how confident we can be with these metrics: AFAIK we only collect container metrics from some specific namespaces.

teemow:
@hervenicol this is for MCs. Imo we collect all metrics on MCs we just exclude other namespaces on WCs. If this is not the case we should fix that imo. I'll take a look...

PR #649, opened 2023-02-14
Add kong rules test

mleisa:
Towards: https://github.com/giantswarm/giantswarm/issues/23854

This PR adds a unit test for `kong.rules.yml`

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/kong.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/kong.rules.yml
@@ -16,7 +16,7 @@ spec:
       annotations:
         description: '{{`Datastore of Kong in namespace {{ $labels.namespace }} is not reachable.`}}'
         opsrecipe: managed-app-kong/#kong-datastore-not-reachable
-      expr: kong_datastore_reachable == 0
+      expr: up{job="kong_datastore_reachable"} == 0
       for: 10m
       labels:
         area: managedservices
--- a/helm/prometheus-rules/templates/alerting-rules/kong.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/kong.rules.yml
@@ -16,7 +16,7 @@ spec:
       annotations:
         description: '{{`Datastore of Kong in namespace {{ $labels.namespace }} is not reachable.`}}'
         opsrecipe: managed-app-kong/#kong-datastore-not-reachable
-      expr: up{job="kong_datastore_reachable"} == 0
+      expr: kong_datastore_reachable == 0
       for: 10m
       labels:
         area: managedservices
--- a/helm/prometheus-rules/templates/alerting-rules/kong.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/kong.rules.yml
@@ -22,7 +22,7 @@ spec:
         area: managedservices
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
-        cancel_if_outside_working_hours: {{ include "workingHoursOnly" . }}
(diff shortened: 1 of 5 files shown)


helm/prometheus-rules/templates/alerting-rules/kong.rules.yml
```diff
@@ -16,7 +16,7 @@ spec:
       annotations:
         description: '{{`Datastore of Kong in namespace {{ $labels.namespace }} is not reachable.`}}'
         opsrecipe: managed-app-kong/#kong-datastore-not-reachable
-      expr: kong_datastore_reachable == 0
```
ubergesundheit, line 19: I have zero knowledge about these expressions, but why does it need to be `up`?

mleisa: `up` is a metric that is always scraped by Prometheus and describes the availability of an instance (1 = available, 0 = unavailable). Since it's always scraped, it would be sensible to use it for alerts like this which just monitor if something is reachable or not. Here's a better explanation: https://prometheus.io/docs/concepts/jobs_instances/

What I'm not sure about is if it is NEEDED and what the Pros/Cons are because the test will pass with just `kong_datastore_reachable == 0` as well (if the test is tweaked). I went with `up` by looking at other rules in this repo and examples online. They seem to exclusively use Prometheus metrics, and `up{}` seems to be the standard for availability metrics.

Maybe @hervenicol could shed some light here.

hervenicol: Well, `kong_datastore_reachable` and `up` metrics are 2 different things.
* `up` is a metric generated by prometheus, that says if a job (= target that prometheus is supposed to scrape) works (= prometheus manages to scrape it) or not. If it has a job `kong_datastore_reachable`, this expression will tell us if it's alive or not.
* But `kong_datastore_reachable` seems to be a metric provided by Kong's prometheus plugin (https://docs.konghq.com/hub/kong-inc/prometheus/). So I doubt `up{job="kong_datastore_reachable"}` is a thing. 
* So, `up{job="name_of_the_kong_job"}` could make sense. But it would not monitor the same thing as `kong_datastore_reachable`: 
    * one checks that the "plugin" can be reached (can be kong, or can be separate). 
    * The other checks that "Datastore is reachable from Kong" according to the doc. What it means / whether kong can work without a datastore depends on Kong internals, I can't tell.

hervenicol: BTW, as kong is not deployed on all clusters, would be nice to mention the clusters you did your tests on.
For instance, I tried both `up{job="kong_datastore_reachable"}` and `kong_datastore_reachable` on gauss, but had no results.

ubergesundheit: thanks for the explanation. You can check anteater/deu01 for a cluster with kong installed for example

Also this means the change needs to be reverted, right?

hervenicol: From what I see in anteater/deu01, yes, this change does not work.
If you want to check if the exporter/plugin is alive, you could probably do something around `up{app="kong-app"}`.
I can't tell you what is the best/right query, it depends what you want and requires kong knowledge.

mleisa: Thanks for the insight Herve, I had missed the fact that `kong_datastore_reachable` is a metric of the Kong plugin. I will revert this change and adjust the test

mcbenjemaa (approved):
Nice test

PR #653, opened 2023-02-27
Add team labels

QuentinBisson:
Towards: https://github.com/giantswarm/giantswarm/issues/18281

This PR adds the team label to prometheus rules so they can be picked up by the change introduced in https://github.com/giantswarm/prometheus-meta-operator/pull/1181

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/_helpers.tpl
+++ b/helm/prometheus-rules/templates/_helpers.tpl
@@ -23,6 +23,7 @@ app.giantswarm.io/branch: {{ .Values.project.branch | replace "#" "-" | replace 
 app.giantswarm.io/commit: {{ .Values.project.commit | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+application.giantswarm.io/team: {{ index .Chart.Annotations "application.giantswarm.io/team" | quote }}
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/_helpers.tpl
```diff
@@ -23,6 +23,7 @@ app.giantswarm.io/branch: {{ .Values.project.branch | replace "#" "-" | replace
 app.giantswarm.io/commit: {{ .Values.project.commit | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+application.giantswarm.io/team: {{ index .Chart.Annotations "application.giantswarm.io/team" | quote }}
```
QuantumEnigmaa, line 26:
```suggestion
application.giantswarm.io/team: {{ index .Chart.Annotations "application.giantswarm.io/team" | default "atlas" | quote }}
```

QuantumEnigmaa (changes requested):

QuantumEnigmaa (approved):

hervenicol (approved):
Looking good.
I wonder if we can override it per rules-file. I mean, the whole chart belongs to us, but in practice each rules file belongs to a different team, right?

QuentinBisson:
Yes we should be able to as long as we have the team. But some rules are for multiple teams :(

PR #665, opened 2023-03-14
Update readme

QuantumEnigmaa:
This PR ...

### Checklist

- [x] Update CHANGELOG.md


QuentinBisson (approved):
Nice i would not update the Changelog though :)

hervenicol (approved):
I was expecting a schema about how inhibitions work, rather than how tests work. That would have shown the the source and target labels and the difference, which we don't see here.

But that will help anyway :)

QuantumEnigmaa:
Maybe I'll add another graph then :)

PR #668, opened 2023-03-15
Add Kyverno deployment not satisfied alert and remove from Honeybadger

fhielpos:
Removing the old Kyverno alert and adding DeploymentNotSatisfied. Also removing it from Honeybadger alert.


stone-z (approved):
LGTM

PR #687, opened 2023-04-05
Check etcd-kubernetes-resources-count-exporter

njuettner:


paurosello (approved):
LGTM, but raised a different question, couldn't we do this with a label filter to make sure that whatever is deployed in there is monitored by default?

whites11 (approved):

njuettner:
@paurosello that is a good idea, we currently don't have a label but I will create an issue for this idea 👍🏻. in the meantime I'll merge it for now.

PR #708, opened 2023-04-21
Add label to trigger inhibition when prometheus-agent down

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/26678

To avoid several types of alerts to fire because metrics are missing, we create an inhibition when prometheus-agent is down.

### Checklist

- [x] Update CHANGELOG.md
- [x] Add [Unit tests](https://github.com/giantswarm/prometheus-rules/#testing)
- [x] Follow [Alert structure](https://github.com/giantswarm/prometheus-rules/#how-alerts-are-structured)
- [ ] Consider [creating a dashboard](https://docs.giantswarm.io/getting-started/observability/grafana/custom-dashboards/) ([guidelines](https://intranet.giantswarm.io/docs/product/ux/guidelines/dashboards/)) (if it does not exist already) to help oncallers monitor the status of the issue.
- [ ] Request review from oncall area, as well as team (e.g: `oncall-kaas-cloud` GitHub group).

PR #709, opened 2023-04-21
README: naming convention for inhibition labels

hervenicol:
After talking with @marieroque about inhibitions for https://github.com/giantswarm/prometheus-meta-operator/pull/1271 and https://github.com/giantswarm/prometheus-rules/pull/708, I think there is something fundamentally wrong with our inhibitions: we should have a convention for source labels names rather than source alert names.

So this PR adds a note about it in the README.

Don't hesitate to discuss this idea.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,10 @@ The `cancel_if_*` labels are used to inhibit alerts, they are defined in [Alertm
 
 The base principle is: if an alert is currently firing with a `source_matcher` label, then all alerts that have a `target_matcher` label are inhibited (or muted).
 
+To make inhibitions easier to read, let's try to follow this naming convention inhibition-related labels:
+* `inhibit_[something]` for `source` matchers
+* `cancel_if_[something]` for `target` matchers
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -85,6 +85,10 @@ The `cancel_if_*` labels are used to inhibit alerts, they are defined in [Alertm
 
 The base principle is: if an alert is currently firing with a `source_matcher` label, then all alerts that have a `target_matcher` label are inhibited (or muted).
 
+To make inhibitions easier to read, let's try to follow this naming convention inhibition-related labels:
+* `inhibit_[something]` for `source` matchers
+* `cancel_if_[something]` for `target` matchers
```
QuentinBisson, line 90: Should we not name this inhibited_by_[something] eventually?

QuentinBisson: I agree and that's why I said "eventually" :D

QuentinBisson (approved):

README.md
QuentinBisson: I just think we should try to move away from the Giant Swarm way of things when open source does it better, has better names

PR #716, opened 2023-04-26
Add a rule to check for Cluster Capi APP failed

primeroz:
Towards: https://github.com/giantswarm/giantswarm/issues/26728

This PR adds a new rule to check for `Cluster CAPI App` in failed state 

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.94.0] - 2023-04-26
+
 ### Added
 
 - Alert `MimirComponentDown` for Mimir app
@@ -1721,7 +1723,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add existing rules from https://github.com/giantswarm/prometheus-meta-operator/pull/637/commits/bc6a26759eb955de92b41ed5eb33fa37980660f2
 
-[Unreleased]: https://github.com/giantswarm/prometheus-rules/compare/v2.93.0...HEAD
+[Unreleased]: https://github.com/giantswarm/prometheus-rules/compare/v2.94.0...HEAD
+[2.94.0]: https://github.com/giantswarm/prometheus-rules/compare/v2.93.0...v2.94.0
 [2.93.0]: https://github.com/giantswarm/prometheus-rules/compare/v2.92.0...v2.93.0
 [2.92.0]: https://github.com/giantswarm/prometheus-rules/compare/v2.91.0...v2.92.0
 [2.91.0]: https://github.com/giantswarm/prometheus-rules/compare/v2.90.0...v2.91.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Alert `MimirComponentDown` for Mimir app
 - Alert `ClusterCapiAppFailed` for Detecting failing `cluster-XXX` apps for CAPI Clusters
 
+### Fixed
+
+- Fix colliding resources names in recording rules
+
 ## [2.93.0] - 2023-04-24
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Alert `MimirComponentDown` for Mimir app
+
 ## [2.93.0] - 2023-04-24
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Alert `MimirComponentDown` for Mimir app
+- Alert `ClusterCapiAppFailed` for Detecting failing `cluster-XXX` apps for CAPI Clusters
 
 ## [2.93.0] - 2023-04-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Alert `MimirComponentDown` for Mimir app
-- Alert `ClusterCapiAppFailed` for Detecting failing `cluster-XXX` apps for CAPI Clusters
+- Alert `ClusterCapiAppFailed` for detecting failing `cluster-XXX` apps representing CAPI workload clusters.
 
 ### Fixed
 
--- a/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/app.rules.yml
@@ -93,7 +93,7 @@ spec:
         topic: releng
     - alert: ClusterCapiAppFailed
       annotations:
-        description: '{{`Cluster CAPI App {{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}, version {{ $labels.version }} is in {{ $labels.status }} state.{{end}}`}}'
+        description: '{{`Cluster CAPI App {{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}, version {{ $labels.version }} is in {{ $labels.status }} state.`}}'
         opsrecipe: app-failed/
       # we have this Frankenstein for namespace checks because go (and prometheus) don't support negative lookaheads in regex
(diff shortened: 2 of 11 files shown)


helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -91,6 +91,22 @@ spec:
...
         topic: releng
+    - alert: ClusterCapiAppFailed
+      annotations:
+        description: '{{`Cluster CAPI App {{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
+        opsrecipe: app-failed/
+      # we have this Frankenstein for namespace checks because go (and prometheus) don't support negative lookaheads in regex
+      expr: label_replace(app_operator_app_info{app=~"cluster-.+", status!~"(?i:(deployed|cordoned|not-installed))", catalog="cluster", team!~"^$|noteam"}, "cluster_id", "$1", "namespace", {{ include "namespaceNotGiantswarm" . }}) == 1
+      for: 30m
+      labels:
+        area: managedservices
```
whites11, line 102: isn't this kaas? this is an app managed by kaas to go alongside a cluster, so I don't think it should be a managed service.
also I think we are missing the `team` label here.

primeroz: I don't know , i copied it from the other `ManagementClusterAppFailed` and `WorkloadClusterappFailed` but sure i can change it :+1:

primeroz: > also I think we are missing the team label here.

I thought the `team` label would come from the actual series ( so clippy for cluster-azure, hydra for cluster-aws and so on )

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -91,6 +91,22 @@ spec:
         cancel_if_outside_working_hours: "true"
         severity: page
         topic: releng
+    - alert: ClusterCapiAppFailed
+      annotations:
+        description: '{{`Cluster CAPI App {{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}, version {{ $labels.version }} is {{if $labels.status }} in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
```
bavarianbidi, line 96:
```suggestion
        description: '{{`Cluster CAPI App {{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}, version {{ $labels.version }} is in {{ $labels.status }} state. {{else}} not installed. {{end}}`}}'
```

according the test-data the `{{if $label.status}}` seems wrong/obsolete?

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
whites11: yeah might be :+1:

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
primeroz: i did run the `make test` and it succeded. commiting your suggestion :+1:

bavarianbidi: yes, the test didn't fail because your test-case was "wrong" (you have two spaces between `is` and `in`)

test/tests/providers/global/app.rules.test.yml
```diff
@@ -100,3 +100,45 @@ tests:
...
+              severity: page
+              sig: none
+              status: failed
+              team: clippy
+              topic: releng
+              upgrade_available: "true"
+              version: 0.0.15
+              version_mismatch: "false"
+            exp_annotations:
+              description: "Cluster CAPI App org-giantswarm/glippy, version 0.0.15 is  in failed state. "
```
bavarianbidi, line 141:
```suggestion
              description: "Cluster CAPI App org-giantswarm/glippy, version 0.0.15 is in failed state. "
```

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
primeroz: ah ! :D

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -91,6 +91,22 @@ spec:
...
+        opsrecipe: app-failed/
+      # we have this Frankenstein for namespace checks because go (and prometheus) don't support negative lookaheads in regex
+      expr: label_replace(app_operator_app_info{app=~"cluster-.+", status!~"(?i:(deployed|cordoned|not-installed))", catalog="cluster", team!~"^$|noteam"}, "cluster_id", "$1", "namespace", {{ include "namespaceNotGiantswarm" . }}) == 1
+      for: 30m
+      labels:
+        area: kaas
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "true"
```
fiunchinho, line 106: Recently, atlas added a new inhibition that _I think_ applies to this use case. It's called `cancel_if_inhibit_prometheus_agent_down`

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -91,6 +91,22 @@ spec:
         cancel_if_outside_working_hours: "true"
         severity: page
         topic: releng
+    - alert: ClusterCapiAppFailed
+      annotations:
+        description: '{{`Cluster CAPI App {{ if $labels.exported_namespace }}{{ $labels.exported_namespace }}{{ else }}{{ $labels.namespace }}{{ end }}/{{ $labels.name }}, version {{ $labels.version }} is in {{ $labels.status }} state.`}}'
+        opsrecipe: app-failed/
+      # we have this Frankenstein for namespace checks because go (and prometheus) don't support negative lookaheads in regex
```
fiunchinho, line 98: It looks like this comment doesn't apply anymore, as it has been extracted into `namespaceNotGiantSwarm`.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Alert `MimirComponentDown` for Mimir app
+- Alert `ClusterCapiAppFailed` for Detecting failing `cluster-XXX` apps for CAPI Clusters
```
fiunchinho, line 15:
```suggestion
- Alert `ClusterCapiAppFailed` for detecting failing `cluster-XXX` apps representing CAPI workload clusters.
```

fiunchinho (commented):
Left a couple of comments, but it's looking good!

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
```diff
@@ -91,6 +91,22 @@ spec:
...
+      expr: label_replace(app_operator_app_info{app=~"cluster-.+", status!~"(?i:(deployed|cordoned|not-installed))", catalog="cluster", team!~"^$|noteam"}, "cluster_id", "$1", "namespace", {{ include "namespaceNotGiantswarm" . }}) == 1
```
QuentinBisson, line 99: The cluster-id is not good here because you will only configure the org-namespace.

I fixed it here https://github.com/giantswarm/prometheus-rules/pull/718/files

Also, do we need an extra alert for this or would it not make sense to add a regex over the catalog in the alert I fixed in my PR?

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
QuentinBisson: Actually it's `cancel_if_prometheus_agent_down`, we need to edit the announcement

helm/prometheus-rules/templates/alerting-rules/app.rules.yml
primeroz: from the discussion on slack (https://gigantic.slack.com/archives/C01176DKNP4/p1682502034772099)  the reason to have an extra alert is that 

* the `cluster-XXX` app is used to create both the MC and WC clusters so it belonged in case only as part of the `ManagementClusterAppFailed` 
* but if i add the `cluster` catalog to the regex of the `ManagementClusterAppFailed` it will also match on the `default-apps-xxxx` which instead belongs **both** to the WorkloadClusters and the ManagementCluster 

So the options where
* merge all the `AppFailed` rules together
* Create a separate alert for Cluster Api APP 

But i am ok with either really :+1:

primeroz: > but if i add the cluster catalog to the regex of the ManagementClusterAppFailed it will also match on the default-apps-xxxx which instead belongs both to the WorkloadClusters and the ManagementCluster

Also just realized i forgot to add `default-apps-.+` to the list

QuentinBisson: I was just thinking that we do not need a distinction between management cluster apps and workload cluster apps because the same team manages both :D

primeroz: for me your #718 is fine , i am gonna close this pr in favour of that one :+1:

QuentinBisson: Is there any other app in the cluster catalogs? Maybe you do not need to app name disctinction

PR #729, opened 2023-05-01
remove heartbeats delay

hervenicol:
Towards: https://github.com/giantswarm/giantswarm/issues/26770

This PR removes the delay before sending heartbeat alerts.
Means as soon as prometheus is up and ready it sends heartbeats, instead of waiting for 10 extra minutes.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix `WorkloadClusterControlPlaneNodeMissing` alerts for all providers (previous fix was not working)
 
+### Changed
+
+- Reduced delay for heartbeats from 10m to none
+
 ## [2.96.0] - 2023-04-28
 
 ### Added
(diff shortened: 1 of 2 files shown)


QuentinBisson:
Why do you think this makes sense? Will we not then bé unaware that prometheus is crashing in a loop?

hervenicol:
> Why do you think this makes sense? Will we not then bé unaware that prometheus is crashing in a loop?

There's 2 different cases why prometheus could be crashing in a loop:
- crashing when reading WAL. In this case, it does not become ready, and does not send heartbeat.
- crashing after being ready, like when receiving data from agent. In this case, the 10min delay is arbitrary, and I'm not even sure we're catching it in a consistent way today. We should create SLO rules for this.

Anyway I'm looking for a series of patches in different directions to reduce "false" heartbeat alarms, and this one looked like the easiest to implement.
Maybe we should do an SLO rule before though.

QuentinBisson:
Sure thing. Then let's try this for a week and check the dashboard showing prometheus availability which is what we should use as an slo in my opinion. If we can reduce oncall alerts for now while keeping the same feature it's fine with me

QuentinBisson (approved):

QuentinBisson:
We cannot release till one prometheus-rule is fixed though

QuentinBisson:
This https://gigantic.slack.com/archives/C020E38NGTZ/p1682676925792619?thread_ts=1682676716.862719&cid=C020E38NGTZ
Should become https://gigantic.slack.com/archives/C020E38NGTZ/p1682677181541959?thread_ts=1682676716.862719&cid=C020E38NGTZ but I did not have time friday to fix it

TheoBrigitte (approved):

PR #731, opened 2023-05-01
added PrometheusAvailabilityRatio alert

hervenicol:
Towards: https://github.com/giantswarm/giantswarm/issues/26770
Related: https://github.com/giantswarm/prometheus-rules/pull/729

This PR creates a new alert, `PrometheusAvailabilityRatio`, that pages during working hours when prometheus was up for less than 80% time in the past 10 hours.

Good to know: The alert is generated from MC prometheus, so if it's dead we have no visibility over it.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
@@ -10,6 +10,25 @@ spec:
   groups:
   - name: prometheus
     rules:
+    - alert: PrometheusAvailabilityRatio
+      annotations:
+        description: '{{`Prometheus {{$labels.pod}} has availability ratio of {{ printf "%.2f" $value }} (min 0.8) over the last 10 hours.`}}'
+        opsrecipe: prometheus-resource-limit-reached/
+      expr: avg(avg_over_time(kube_pod_status_ready{namespace=~"(.*)-prometheus", condition="true"}[10h])) by (pod) < 0.8
+      # At startup, availability starts at 0 for a few minutes. So ratio grows slowly from 0.
+      for: 30m
+      labels:
+        area: empowerment
+        cancel_if_any_apiserver_down: "true"
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_has_no_workers: "true"
+        cancel_if_outside_working_hours: "true"
+        dashboard: promavailability/prometheus-availability
+        severity: page
+        team: atlas
+        topic: observability
     - alert: PrometheusCantCommunicateWithKubernetesAPI
       annotations:
         description: '{{`Prometheus can''t communicate with Kubernetes API.`}}'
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
```diff
@@ -10,6 +10,25 @@ spec:
   groups:
   - name: prometheus
     rules:
+    - alert: PrometheusAvailabilityRatio
```
QuentinBisson, line 13: I guess this is a temporary alert until we have an slo alert right? Can you add an ops-recipe link even if it's empty?

hervenicol: Well, I put the `prometheus-resource-limit-reached` opsrecipe because it explains quite well how to understand and fix all the reasons why prometheus would die.
That's better than an empty opsrecipe I guess?

hervenicol: Also, regarding SLOs:
I could have found a way to have an "SLO burn rate" alert. But I don't like them for a few reasons:
- they are quite generic and point to a generic opsrecipe
- the promql query is harder to understand

As I already had a nice query for this ratio in the availability dashboard, I thought I could as well use it.

Proper solution would be to use SLOTH, but it's not ready yet.

QuentinBisson: I agree with all your points, because even if we have an slo alert, each should have their different opsrecipe. I was thinking the expression behind your alert would be a good first step for our SLO

helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
```diff
@@ -23,7 +42,7 @@ spec:
         cancel_if_cluster_status_updating: "true"
         cancel_if_cluster_status_deleting: "true"
         cancel_if_cluster_has_no_workers: "true"
-        cancel_if_outside_working_hours: {{ include "workingHoursOnly" . }}
+        cancel_if_outside_working_hours: "false"
```
marieroque, line 45: Why this change ?

QuentinBisson (approved):
LGTM for a first attempt

helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
hervenicol: Short answer: The "workingHoursOnly" variable is very counter-intuitive and should not be used.

Longer answer:
It's defined here: https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/_helpers.tpl#L44-L50
It's set to `false` on most installations,
except for `openstack` and `capz` installations where it's set to `true`.
This kind of inconsistency is a nice recipe for hard-to-debug alerts :shrug:

TheoBrigitte (approved):

helm/prometheus-rules/templates/alerting-rules/prometheus.rules.yml
marieroque: I agree.
This change affects the actual behaviour for capz right ?
I don't know why we have a specific case for capz BTW...

hervenicol: > This change affects the actual behaviour for capz right ?

Right.

> I don't know why we have a specific case for capz BTW...

Not sure there's a valid reason :shrug:

PR #752, opened 2023-05-08
Update Grafana cloud rules for Kyverno policies

stone-z:
Towards: https://github.com/giantswarm/giantswarm/issues/26857

This PR includes additional Pod controller types to have their policy report results shipped to grafana cloud.
It also explicitly limits to only shipping reports from MC components.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
+++ b/helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
@@ -236,13 +236,13 @@ spec:
     # Kyverno-related resource counts by kind
     - expr: sum(etcd_kubernetes_resources_count{kind=~".*.kyverno.io|clusterpolicyreports.wgpolicyk8s.io|policyreports.wgpolicyk8s.io"}) by (cluster_id, kind)
       record: aggregation:kyverno_resource_counts
-    # Kyverno policy status by team
+    # Kyverno policy status by team - Deployments
     - expr: |-
         sum(
           label_join(policy_report_result{
             policy!="check-deprecated-apis-1-25",
             cluster_type="management_cluster",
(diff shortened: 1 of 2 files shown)


stone-z:
Ok this is hideous but works. The joins and relabeling to try to force this into a single metric was beyond me -- if any monitoring guru has a suggestion for a nice way to just get a `type="deployment|daemonset|etc"` label added here without some really nasty promql, I'm all over it.
But for the time being, this will at least get the metrics we need

stone-z:
Tested on `gremlin` and already started reworking global dashboards to use the new metrics

helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
```diff
@@ -236,12 +236,13 @@ spec:
     # Kyverno-related resource counts by kind
     - expr: sum(etcd_kubernetes_resources_count{kind=~".*.kyverno.io|clusterpolicyreports.wgpolicyk8s.io|policyreports.wgpolicyk8s.io"}) by (cluster_id, kind)
       record: aggregation:kyverno_resource_counts
-    # Kyverno policy status by team
+    # Kyverno policy status by team - Deployments
```
QuentinBisson, line 239: Maybe @hervenicol or @TheoBrigitte have a better solution in mind.

I would be careful about noteam though. We might want to add a metric in grafana cloud to know if we have apps without teams

QuentinBisson: Yes I meant that for my team :)

stone-z:
I'd like to get this out so we can start getting data. Will merge + release sometime today and we can optimize later if somebody has an idea

hervenicol (approved):
Your query looks great to me. :heart: 
I wouldn't do better.

PR #757, opened 2023-05-12
reduce PrometheusRuleFailures interval to 5m

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/26932

This PR changes the range interval from 30m to 5m in PrometheusRuleFailures alert.

This has 2 implications :

- Quicker aler resolution, has alert will resolve/close 5m after the issue is gone.
- Less noise:
    As 30m range query with for:1h means we get paged whenever there are at least 1 error every 30minutes for 1h. (very sensitive)
    As 5m range query with for:1h means we get paged whenever there are at least 1 error every 5mn for 1h.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


hervenicol (approved):
:heart:

PR #786, opened 2023-06-14
Containerd low volume

njuettner:
Containerd Volume has never been checked since we moved customer / giantswarm pods to containerd

This adds a alert if the containerd volume has less than 10% available.

Towards: https://github.com/giantswarm/giantswarm/issues/27015

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/disk.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/disk.management-cluster.rules.yml
@@ -27,7 +27,7 @@ spec:
       annotations:
         description: '{{`Containerd volume {{ $labels.mountpoint}} on {{ $labels.instance }} does not have enough free space.`}}'
         opsrecipe: low-disk-space/#containerd-volume
-      expr: node_filesystem_free_bytes{mountpoint=~"(/rootfs)?/var/lib/containerd"} < (10 * 1024 * 1024 * 1024)
+      expr: 100 * node_filesystem_free_bytes{mountpoint=~"(/rootfs)?/var/lib/containerd"} / node_filesystem_size_bytes{mountpoint=~"(/rootfs)?/var/lib/containerd"} < 10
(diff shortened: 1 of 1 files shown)


primeroz (commented):
if we have unit test for the other disk alerts we should probably add one for this as well

bavarianbidi:
> I'd strongly suggest you to create a SLO instead of a plain prometheusrule for this : [giantswarm/sloth-rules](https://github.com/giantswarm/sloth-rules)

i'm not against doing it that way, but i want to understand when should we do SLO based alerts?
hat's the SLO we want to have here? _We guarantee, that the disk is never filled up to X%_ If so, we have to implement much more restrict eviction policies that it's technically very hard to fill this disk space - which might has other impact.

Few days ago i've alread started a slack-thread about when to use `slo` alerts and when to use non-SLO alerts (as some components need a manual check or further debugging but the availability isn't affected) 
https://gigantic.slack.com/archives/C0E0V16DC/p1684932308297269

 @QuantumEnigmaa would be great to get your opinion here as well :pray:

QuantumEnigmaa:
> when should we do SLO based alerts?

I would say that if the alert is about availability (as it is the case here) then it  would be best to create SLO based alerts for it as it uses a multi-window method. This means that we can tune the SLO so that we only get "warned" as long as the availability doesn't drop below a certain percentage but also get paged for another amount defined in advance. 

I also think that some alerts would best stick with plain prometheusrules alerts but I don't have specific examples in mind.
We should have a global discussion among SIG Monitoring or even wider about this topic so that we can define strict guidelines for when not to use SLO based alerts (since most of the alerts will most likely fall in the "should be a SLO" category).

In any case, for this particular alert concerning disk space, an SLO based alert (with a 90% objective from what I see) is really the way to go in my opinion.

QuantumEnigmaa (approved):

PR #788, opened 2023-06-14
Fix description for alert to use the actual label for cluster name

primeroz:
Right now we are missing the cluste rname into the alert

image

because the label is `name` and not `cluster_name`

image

### Checklist

- [x] Update CHANGELOG.md


QuentinBisson (approved):

primeroz:
we did have the unit test ... but the label in the unit test was set to `cluster_name` rather than the real label on the metric. 

so the unit test succeded but it was still wrong :)

PR #826, opened 2023-07-11
remove-CoreDNSLatencyTooHigh

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/27506

we tried enabling this alert, but it pages all the time and so it doesn't work well.
We are working on an SLO alert (sloth based) to measure this problem


silenteh (approved):

njuettner (approved):
👍🏻

PR #829, opened 2023-07-12
Add Kyverno rules

fhielpos:
Towards: https://github.com/giantswarm/giantswarm/issues/27333

Still a WIP, I am working on adding the opscrecipes

### DIFF
--- a/test/hack/checkLabels/go.mod
+++ b/test/hack/checkLabels/go.mod
@@ -5,7 +5,7 @@ go 1.19
 require (
 	// Try to keep version in sync with our prometheus rule CRD version.
 	// see https://github.com/giantswarm/prometheus-operator-crd/blob/master/helm/prometheus-operator-crd/Chart.yaml#L11
-	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.66.0
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.67.1
 	sigs.k8s.io/yaml v1.3.0
 )
 
--- a/test/hack/checkLabels/go.sum
+++ b/test/hack/checkLabels/go.sum
@@ -499,6 +499,8 @@ github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.65.2 h
 github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.65.2/go.mod h1:xcfWyzl4BpEe5jnVkw7D1yCHU7GHjfjCERJsEfGbpSU=
 github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.66.0 h1:PPW01FLVjJHMNcbAL1DDD9EZceSQKMOU/VpK0irrxrI=
 github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.66.0/go.mod h1:KZHvrby65G+rA4V/vMTUXDV22TI+GgLIrCigYClpjzk=
+github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.67.1 h1:u1Mw9irznvsBPxQxjUmCel1ufP3UgzA1CILj7/2tpNw=
+github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.67.1/go.mod h1:KZHvrby65G+rA4V/vMTUXDV22TI+GgLIrCigYClpjzk=
 github.com/prometheus/alertmanager v0.22.2 h1:JrDZalSEMb2/2bqGAhls6ZnvOxbC5jMIu29JV+uWTC0=
 github.com/prometheus/alertmanager v0.22.2/go.mod h1:rYinOWxFuCnNssc3iOjn2oMTlhLaPcUuqV5yk5JKUAE=
 github.com/prometheus/alertmanager v0.25.0 h1:vbXKUR6PYRiZPRIKfmXaG+dmCKG52RtPL4Btl8hQGvg=


fhielpos:
@Strigix , your rule seems to have a syntax error when trying to evaluate the tests. It seems that `&&` is not admitted here.

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+  creationTimestamp: null
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+  name: kyverno.all.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+  - name: webhooks
+    rules:
+    - alert: KyvernoWebhooksPresent
```
stone-z, line 13:
```suggestion
    - alert: KyvernoWebhookHasNoAvailableReplicas
```

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+      annotations:
+        description: 'Kyverno has no available replicas but webhooks are present.'
```
stone-z, line 15: It would be nice to include the actual webhook name in the description if that info is available in the alert

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
stone-z: I also wonder if this alert could be made generic to any webhook. Might require quite some promql magic though

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+        opsrecipe: kyverno-webhooks/
+      expr: sum(kube_validatingwebhookconfiguration_info{validatingwebhookconfiguration=~"kyverno-.*"}) > 0 and sum(kube_deployment_status_replicas{deployment=~"kyverno|kyverno-admission-controller"}) == 0
```
stone-z, line 17: Conscious decision: let's wait and see if we hit a case where we need to also monitor `mutatingwebhookconfigurations`. I expect that cases where there's a mutating webhook and no validating webhook are virtually nonexistent

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "false"
+        severity: page
+        team: shield
+        topic: kyverno
+  - name: reports
+    rules:
+    - alert: KyvernoReportsTooHigh
+      annotations:
+        description: '{{`Kyverno {{ $labels.kind }} are too high.`}}'
```
stone-z, line 32:
```suggestion
        description: '{{`Kyverno {{ $labels.kind }} are too high. This is an indicator that Kyverno's report processing may not be keeping up with cluster demand.`}}'
```

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+      labels:
+        area: managedservices
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "true"
+        severity: notify
+        team: shield
+        topic: kyverno
+    - alert: KyvernoWronglyScaled
```
stone-z, line 62:
```suggestion
    - alert: KyvernoUnsafeReplicaCount
```

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+      annotations:
+        description: 'Kyverno deployment scaling has an unsupported amount of replicas'
```
stone-z, line 64:
```suggestion
        description: 'Kyverno's admission controller deployment must use at least 3 replicas, or be scaled to 0.'
```

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+        opsrecipe: KyvernoWronglyScaled/
+      expr: sum(kube_deployment_spec_replicas{deployment="kyveno"}) > 0 && 3 <
```
stone-z, line 66: Probably something like 0 < sum < 3 was meant here? I'd hope that's valid promql syntax

fhielpos: Not sure @Strigix , where you able to check this one?

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
fhielpos: We have it in the alert, I think the problem is that there are two different labels depending if it is validating or mutating, plus if you have 10 webhooks then the alert would be super crowded.

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+        description: '{{`Kyverno {{ $labels.kind }} are too high.`}}'
+        opsrecipe: kyverno-reports/
+      expr: aggregation:kyverno_resource_counts{kind=~"(clusteradmissionreports|admissionreports).kyverno.io"} > 50000
+      for: 15m
+      labels:
+        area: managedservices
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "false"
```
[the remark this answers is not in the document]
fhielpos: I think it could be possible, the thing is if this happens during a weekend/holiday then we could avoid issues before they happen. We could also increase the threshold that would slow the alert from triggering.

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+        team: shield
+        topic: kyverno
+  - name: replicas
+    rules:
+    - alert: KyvernoScaledDownTooLong
+      annotations:
+        description: 'Kyverno has been scaled down for too long.'
+        opsrecipe: kyverno-scaled-down/
+      expr: sum(kube_deployment_spec_replicas{deployment="kyverno"}) == 0
+      for: 1h
```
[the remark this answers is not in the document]
fhielpos: Ah yeah, it was a random value I chose. We can have like 5h or so and leave just for BH.

fhielpos:
Closes: https://github.com/giantswarm/giantswarm/issues/27834

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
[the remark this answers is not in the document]
fhielpos: I aligned with other's alerts and set it to 4h, it should never page more than 2x per BH shift (but if it even alerts once then we should create a silence for it if it's expected).

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -0,0 +1,76 @@
...
+spec:
+  groups:
+  - name: webhooks
+    rules:
+    - alert: KyvernoWebhookHasNoAvailableReplicas
+      annotations:
+        description: 'Kyverno has no available replicas but webhooks are present.'
+        opsrecipe: kyverno-webhooks/
+      expr: sum(kube_validatingwebhookconfiguration_info{validatingwebhookconfiguration=~"kyverno-.*"}) > 0 and sum(kube_deployment_status_replicas{deployment=~"kyverno|kyverno-admission-controller"}) == 0
+      for: 5m
```
stone-z, line 18: On a second look, 5m might be too noisy for an out of business hours page. We can try this and adjust if it starts waking folks up for silly reasons

stone-z (approved):

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
fhielpos: Ooooooh you are right

fhielpos: Well, but also this is something that could break the cluster. Maybe we could set it to 15/30m?

PR #836, opened 2023-07-19
 Add `DNSRequestDurationTooSlow` to catch slow DNS.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/27506

add new alert that fires when 99th percentile of coredns name resolution takes more than 1s for 15m straight.

Note: we have a similar SLO alert for DNS latency, but that is BH only and we would like to have this temporary alert until we promote the other one to 24h alert


calvix (approved):

njuettner (approved):

hervenicol (approved):
LGTM

mcharriere (approved):

PR #903, opened 2023-09-12
Change ClusterUnhealthyPhase severity to page

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/28000

On this PR we are changing the severity of `ClusterUnhealthyPhase` from `notify` to `page`. We'd like to this change so that we get notified when a cluster is not behaving properly, or it's stuck `Deleting`.

### Checklist

- [X] Update CHANGELOG.md


helm/prometheus-rules/templates/alerting-rules/capi-cluster.rules.yml
```diff
@@ -15,7 +15,7 @@ spec:
           labels:
```
marieroque, line 15: What about other valid phases like Provisioning, Running, Pending... ?

fiunchinho: All are valid phases as long as the cluster is not stuck on one of them for too long. The only phase that the cluster should be for long time is the `Provisioned` phase. So the alert will page if the cluster is for 1h on a phase that is not `Provisioned`.

calvix (approved):

AndiDog (approved):
I didn't find any recent such alerts in chat (`#alert-phoenix{,-test}`), so changing this should have no immediate impact, right?

fiunchinho:
> I didn't find any recent such alerts in chat (`#alert-phoenix{,-test}`), so changing this should have no immediate impact, right?

We should've received messages in slack, but I forced a cluster to be stuck `Deleting` to trigger it. But I think there is something wrong with installations labeled as `stable-testing` and their alerting. I've opened this yesterday https://github.com/giantswarm/giantswarm/issues/28106

marieroque (approved):

erkanerol (approved):

PR #912, opened 2023-09-19
split ksm alerts in 2 separate ones

QuantumEnigmaa:
This PR splits the KSMDown alert into 2 different alerts : 
- the KSMDown one which triggers whenever the KSM component is actually down
- the KSMNotRetrievingMetrics one which triggers when KSM is not able to retrieve metrics from the cluster

### Checklist

- [x] Update CHANGELOG.md


helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
```diff
@@ -76,3 +71,24 @@ spec:
         severity: page
         team: atlas
         topic: observability
+
+    - alert: KubeStateMetricsNotRetrievingMetrics
```
QuentinBisson, line 75: What's the difference with KubeSecretMetricMissing and KubeStateMetricsSlow?

QuentinBisson: Also we have a KSMdown inhibition :)

QuantumEnigmaa: From my own understanding, `KubeStateMetricsSlow` is monitoring the response time from KSM to make sure it doesn't take too long to retrieve metrics while `KubeStateMetricsNotRetrievingMetrics` is making sure that there are actually metrics retrieved by KSM

But maybe we can consider that if KSM is taking too long to retrieve metrics then it means that it's unable to retrieve it and thus we can get rid of `KubeStateMetricsNotRetrievingMetrics` to only rely on `KubeStateMetricsSlow`

QuentinBisson: Sure about the slow one but is this alert not the same as those https://github.com/giantswarm/prometheus-rules/blob/737a5e989ff13a86f868eb825ea0e7d71e8842fd/helm/prometheus-rules/templates/alerting-rules/kube-state-metrics.rules.yml#L31?

Also maybe we should regroup all KSM related alerts to this file

helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
```diff
@@ -58,11 +58,6 @@ spec:
           # vintage clusters without servicemonitor
           label_replace(up{app="kube-state-metrics",container=""}, "ip", "$1.$2.$3.$4", "node", "ip-(\\d+)-(\\d+)-(\\d+)-(\\d+).*") == 0 or absent(up{app="kube-state-metrics",container=""} == 1)
         )
-        or
```
QuentinBisson, line 61: Could you also move the KSM down alert?

helm/prometheus-rules/templates/alerting-rules/up.all.rules.yml
QuantumEnigmaa: Not sure, is `kube_configmap_created` a random metric normally retrieved by KSM that we check to make sure KSM is retrieving metrics in general ? 

In that case I'd prefer also keeping the new alert because its name is more straightforward about its usage and moreover it indicates that KSM as a whole is not able to retrieve metrics while the `KubeConfigMapCreatedMetricMissing` would indicate that KSM is not able to retrive metrics from a particular instance

QuentinBisson: Well kube_configmap_created is the metric exposed by KSM when access the apiserver but we can have both if you prefer

QuantumEnigmaa: I guess it would make sense to have both yes

helm/prometheus-rules/templates/alerting-rules/kube-state-metrics.rules.yml
```diff
@@ -28,6 +56,27 @@ spec:
...
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_has_no_workers: "true"
+        inhibit_kube_state_metrics_down: "true"
+        cancel_if_kubelet_down: "true"
+        cancel_if_kube_state_metrics_down: "true"
+        cancel_if_outside_working_hours: "false"
+        severity: page
+        team: atlas
+        topic: observability
```
hervenicol, line 79: When this condition was added, we wanted it to inhibit KSM-dependent alerts.
But I think, if it takes 60min to fire, it won't prevent KSM-dependent alerts from firing.

If we're afraid it would be too noisy, we can try setting it to not page outside of business hours. WDYT?

helm/prometheus-rules/values.yaml
```diff
@@ -2,8 +2,8 @@ name: prometheus-rules
 namespace: monitoring
 serviceType: managed
 project:
-  branch: '[[ .Branch ]]'
-  commit: '[[ .SHA ]]'
+  branch: 'split-ksm-alerts'
+  commit: 'b67266c8479ebef0cdd918fa9fc5e44717edf794'
```
hervenicol, line 6: I think this was added / commited by mistake.

helm/prometheus-rules/templates/alerting-rules/kube-state-metrics.rules.yml
```diff
@@ -10,6 +10,34 @@ spec:
...
+          label_replace(up{app="kube-state-metrics",container=""}, "ip", "$1.$2.$3.$4", "node", "ip-(\\d+)-(\\d+)-(\\d+)-(\\d+).*") == 0 or absent(up{app="kube-state-metrics",container=""} == 1)
+        )
+      for: 15m
+      labels:
+        area: kaas
+        cancel_if_apiserver_down: "true"
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_has_no_workers: "true"
+        inhibit_kube_state_metrics_down: "true"
```
AndiDog, line 34: Does this inhibition make sense for `KubeStateMetricsDown` itself?

hervenicol: Inhibitions work with a `source` label and a `target` label: if an alert with the `source` label fires, alerts with the `target` label are inhibited.

Labels `inhibit_xxx` are `source` labels. This basically means "when this alert fires, please inhibit alerts depending on `kube_state_metrics_down`.

helm/prometheus-rules/Chart.yaml
```diff
@@ -4,7 +4,7 @@ engine: gotpl
 home: https://github.com/giantswarm/prometheus-rules
 icon: https://s.giantswarm.io/app-icons/1/png/default-app-light.png
 name: prometheus-rules
-appVersion: '[[ .AppVersion ]]'
+appVersion: '[[ .Version ]]'
```
hervenicol, line 7: Is it an expected change?
Maybe it deserves a word in PR description and in changelog, as users may see an inconsistency in versions.

QuantumEnigmaa: No this came with merging conflicts with the master branch

helm/prometheus-rules/templates/alerting-rules/kube-state-metrics.rules.yml
QuantumEnigmaa: I guess we can both set it not to page outside of business hours and reduce its time to page to 30min

hervenicol: It's been changed to:
- 20min
- cancel_if_outside_working_hours

Fine with me :smiley:

hervenicol (approved):
LGTM

whites11 (approved):
If atlas is happy I am happy

T-Kukawka (approved):

PR #913, opened 2023-09-19
Fix prometheus agent alerts

QuentinBisson:

---
Towards: https://github.com/giantswarm/...

This PR removes the outside business hours inhibition from the PrometheusAgent inhibiting alerts because otherwise, well, the inhibition is useless

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/prometheus-agent.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/prometheus-agent.rules.yml
@@ -38,7 +38,6 @@ spec:
         cancel_if_cluster_is_not_running_prometheus_agent: "true"
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
-        cancel_if_outside_working_hours: "true"
     ## Page Atlas if prometheus agent is missing shards to send samples to MC prometheus.
     - alert: PrometheusAgentShardsMissing
       annotations:
@@ -74,5 +73,4 @@ spec:
         cancel_if_cluster_is_not_running_prometheus_agent: "true"
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
-        cancel_if_outside_working_hours: "true"
 {{- end }}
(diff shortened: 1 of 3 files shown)


QuantumEnigmaa:
Is the default value for `cancel_if_outside_working_hours` false ? Otherwise instead of removing the label we should keep it and set it to false right ?

QuentinBisson:
The default value is that it is not set. I'd rather we remove it altogether to clean up the alert and avoid possible mistakes in changing that value

helm/prometheus-rules/templates/alerting-rules/prometheus-agent.rules.yml
```diff
@@ -74,5 +73,4 @@ spec:
         cancel_if_cluster_is_not_running_prometheus_agent: "true"
         cancel_if_cluster_status_creating: "true"
         cancel_if_cluster_status_deleting: "true"
-        cancel_if_outside_working_hours: "true"
```
hervenicol, line 77: This one is quite young and can be very noisy, so I'm not sure we should unleash it so soon.
It's been fixed on last Friday, and we immediately had to silence it on clusters that use heavy autoscaling.

QuentinBisson: Sure so https://github.com/giantswarm/prometheus-rules/pull/911/files should have fixed the paging issues that happened during the night. We need to rework it a bit but we cannot not be paged when a Prometheus Agent is down though

hervenicol: Sure.
So I think for now it would be best to have:
- `PrometheusAgentFailing` pages at night
- `PrometheusAgentShardsMissing` keeps it quiet at night (for now)

QuentinBisson: Latest commit should address that until we fixed the shard computation

hervenicol (approved):

PR #945, opened 2023-11-08
Add Kyverno updaterequests alert and tests

fhielpos:
Closes: https://github.com/giantswarm/giantswarm/issues/28807
Signed-off-by: Franco <franco@giantswarm.io>


helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
```diff
@@ -42,6 +42,21 @@ spec:
         severity: page
         team: shield
         topic: kyverno
+    - alert: KyvernoUpdateRequestsCountTooHigh
+      annotations:
+        description: "{{`Kyverno {{ $labels.kind }} are too high. This is an indicator that Kyverno\'s background controller may not be able to create some resources.`}}"
```
QuentinBisson, line 47: Why did you add a \ here ? Otherwise LGTM

QuentinBisson (approved):

helm/prometheus-rules/templates/alerting-rules/kyverno.all.rules.yml
fhielpos: Not sure, I copied it from another rule, will try without it

PR #959, opened 2023-11-16
send used images to grafana cloud

QuentinBisson:
Before adding a new alerting rule into this repository you should consider creating an SLO rules instead.
SLO helps you both increase the quality of your monitoring and reduce the alert noise.

* How to create a SLO rule: https://github.com/giantswarm/sloth-rules#how-to-create-a-slo
* Documentation: https://intranet.giantswarm.io/docs/monitoring/slo-alerting/

---
Towards: https://github.com/giantswarm/roadmap/issues/2913

This PR lists the images used in our clusters so it's easier for us to cleanup those we do not use

### Checklist

- [ ] Update CHANGELOG.md
- [ ] Add [Unit tests](https://github.com/giantswarm/prometheus-rules/#testing)
- [ ] Follow [Alert structure](https://github.com/giantswarm/prometheus-rules/#how-alerts-are-structured)
- [ ] Consider [creating a dashboard](https://docs.giantswarm.io/getting-started/observability/grafana/custom-dashboards/) ([guidelines](https://intranet.giantswarm.io/docs/product/ux/guidelines/dashboards/)) (if it does not exist already) to help oncallers monitor the status of the issue.
- [ ] Request review from oncall area, as well as team (e.g: `oncall-kaas-cloud` GitHub group).

PR #960, opened 2023-11-17
Add two aggregations about images from Docker Hub

marians:
Towards https://github.com/giantswarm/roadmap/issues/2882

Adds two aggregations to track usage of Docker Hub:

- `aggregation:docker:containers_using_dockerhub_image`: Number of containers running with image from docker.io.
- `aggregation:docker:containers_using_dockerhub_image_relative`: Percentage of all running containers that use an image from docker.io (range 0.0 to 1.0).

### DIFF
--- a/helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
+++ b/helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
@@ -105,6 +105,10 @@ spec:
     rules:
     - expr: sum(engine_daemon_image_actions_seconds_count) by (cluster_id, cluster_type, action)
       record: aggregation:docker:action_count
+    - expr: sum(kube_pod_container_info{image=~"docker\\.io/.*"})
+      record: aggregation:docker:containers_using_dockerhub_image
+    - expr: sum(kube_pod_container_info{image=~"docker\\.io/.*"}) / sum(kube_pod_container_info{})
+      record: aggregation:docker:containers_using_dockerhub_image_relative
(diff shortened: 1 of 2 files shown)


piontec (approved):
Nice!
Can we also get percentage of all the registries used, so we know when old ones vs the new one are used?

marians:
> Can we also get percentage of all the registries used, so we know when old ones vs the new one are used?

With `sum(kube_pod_container_info{image=~"docker\\.io/.*"}) / sum(kube_pod_container_info{})` we get the share of containers from Docker Hub as a percentage of all containers. I'm not sure how we could really count unique registries.

helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
```diff
@@ -105,6 +105,10 @@ spec:
     rules:
     - expr: sum(engine_daemon_image_actions_seconds_count) by (cluster_id, cluster_type, action)
       record: aggregation:docker:action_count
```
QuentinBisson, line 107: @marians could you also add init containers here ? I think the metric is named kube_pod_init_container_info

QuentinBisson: I am using the same here https://github.com/giantswarm/prometheus-rules/pull/959/

marians: Very good suggestion! These weren't on my radar as an extra metric. Will make an extra PR.

PR #974, opened 2023-11-27
Check if KSM was up 2 minutes ago in WorkloadClusterCriticalPodNotRunningAWS alert

weseven:
Towards: https://github.com/giantswarm/giantswarm/issues/27895

This PR adds a condition to check if the Kube State Metrics pod was up 2 minutes before this query, in order to prevent triggering of the alert when the KSM was down (with the absence of metrics triggering the second half of the alert) and just came up (lifting the alert inhibition, but potentially triggering this false alert). 

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - reduced sensitivity for lokiringunhealthy
+- added check if KSM was up 2 minutes ago in `WorkloadClusterCriticalPodNotRunningAWS` alert
 
 ## [2.144.0] - 2023-11-27
 
--- a/helm/prometheus-rules/templates/alerting-rules/aws.workload-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/aws.workload-cluster.rules.yml
@@ -32,7 +32,7 @@ spec:
       annotations:
         description: '{{`Critical pod {{ $labels.namespace }}/{{ $labels.pod }} is not running.`}}'
         opsrecipe: critical-pod-is-not-running/
-      expr: kube_pod_container_status_running{namespace="kube-system",container=~"(k8s-api-server|k8s-controller-manager|k8s-scheduler)"} != 1 or label_replace(absent(kube_pod_container_status_running{namespace="kube-system",container="k8s-api-server"}), "pod", "$1", "container", "(.+)") == 1 or label_replace(absent(kube_pod_container_status_running{namespace="kube-system",container="k8s-controller-manager"}), "pod", "$1", "container", "(.+)") == 1 or label_replace(absent(kube_pod_container_status_running{namespace="kube-system",container="k8s-scheduler"}), "pod", "$1", "container", "(.+)") == 1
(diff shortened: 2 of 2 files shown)


hervenicol:
Thanks for looking into this!

Here is a first feedback:

* `offset` will take an instant time in the past, right? What if it was up 5min ago, but had a glitch between 4 and 3 minutes ago? Maybe something like `min_over_time` would be safer?
* the `ksmdown` alert is supposed to check that KSM is up and running, and triggers an inhibition that will prevent this alert from paging. Maybe we should rather act on increasing the inhibition time window?

Actually we (Atlas) were talking about this alert this morning, and we thought we should split the `ksmdown` alert in 2:
- one that fires the actual alert, that would not change
- one that fires the inhibition alert, that would last longer

I did not check how easy/hard that would be though.
Happy to discuss it with you if you want to take over or have ideas.

weseven:
@hervenicol Thanks for reviewing!
You're absolutely right, in case the KSM is flapping using up with offset might be misleading.
I think what you are suggesting (having the alert that controls the inhibitions be a bit more "sticky", that is, not lifting the inhibition immediately) is the better way to approach this. 
That was something I was also thinking about as it's the proper place to fix for all the alerts, but wasn't sure how to approach that since there are already the `KubeStateMetricsSlow` and `KubeStateMetricsNotRetrievingMetrics` that possibly address similar use cases.
Also I'm not sure if it would help (or worsen things) to reduce the time windows for those two alerts; I guess we'll need to examine real world occurrences of the issue, although (fortunately) I don't see this happening recently.

weseven:
>  although (fortunately) I don't see this happening recently.
I had to write that down, didn't I...
It happened yesterday: https://gigantic.slack.com/archives/C067ZGZ6XNY

Seems we are in the exact case where the KSM was down for a bit (~21 minutes), came back up, lifted the inhibition and immediately the WorkloadClusterCriticalPodNotRunningAWS triggered even though those pods were ok: https://opsg.in/a/i/giantswarm/6b1a8c26-396a-49bf-a100-35ceedc57b62-1701293882325

Some metrics on grafana:

[up{app="kube-state-metrics", cluster_id="pcn01"}](https://grafana.g8s.argali.pek.aws.k8s.adidas.com.cn/explore?panes=%7B%22y5E%22:%7B%22datasource%22:%22PBFA97CFB590B2093%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22up%7Bapp%3D%5C%22kube-state-metrics%5C%22,%20cluster_id%3D%5C%22pcn01%5C%22%7D%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22PBFA97CFB590B2093%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%221701291600000%22,%22to%22:%221701295200000%22%7D%7D%7D&schemaVersion=1&orgId=1)

[kube_pod_container_status_running{container=~"(k8s-api-server|k8s-controller-manager|k8s-scheduler)",namespace="kube-system", cluster_id="pcn01"}](https://grafana.g8s.argali.pek.aws.k8s.adidas.com.cn/explore?panes=%7B%22y5E%22:%7B%22datasource%22:%22PBFA97CFB590B2093%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22kube_pod_container_status_running%7Bcontainer%3D~%5C%22%28k8s-api-server%7Ck8s-controller-manager%7Ck8s-scheduler%29%5C%22,namespace%3D%5C%22kube-system%5C%22,%20cluster_id%3D%5C%22pcn01%5C%22%7D%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22PBFA97CFB590B2093%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%221701291600000%22,%22to%22:%221701295200000%22%7D%7D%7D&schemaVersion=1&orgId=1)


Metrics seem to come back at 22:37:15.
I don't know if it's relevant, but looking at the "up" metric, it seems the "metric" endpoint (port 8081) of KSM was discovered up at 22:36:30 ~45s before the http endpoint (port 8080).

hervenicol:
Is this PR still relevant @weseven?

weseven:
I think not; we can close this down, if the same issue catches my eye again we'll start from here :)

PR #986, opened 2023-12-07
Update AlertManagerNotificationsFailing routing on Atlas slack for opsgenie integration

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/28986

### Checklist

- [x] Update CHANGELOG.md

### How to test this PR

1. scale down prometheus-meta-operator
2. edit `alertmanager-config` secret by setting a wrong opsgenie key
3. you can check quickly the alert rule is working `alertmanager_notifications_failed_total{integration="opsgenie"}`. You should have result > 0
4. wait for `AlertmanagerNotifyNotificationsFailing` firing in [#alert-atlas-test](https://gigantic.slack.com/archives/C04AT72MKNK)


helm/prometheus-rules/templates/alerting-rules/alertmanager.rules.yml
```diff
@@ -15,21 +15,45 @@ spec:
...
         severity: page
         team: atlas
         topic: monitoring
+        cancel_if_outside_working_hours: "true"
+    - alert: AlertManagerOpsgenieNotificationsFailing
+      annotations:
+        description: '{{`AlertManager {{ $labels.integration }} notifications are failing.`}}'
+        opsrecipe: alert-manager-notifications-failing/
+      expr: rate(alertmanager_notifications_failed_total{cluster_id!~"akita|argali|giraffe", integration="opsgenie"}[5m]) > 0
+      for: 2m
```
marieroque, line 31: I set to 2m because when an alert fails to notify opsgenie, it increases just by one `alertmanager_notifications_failed_total` counter. 
But the rate over 5m fails down to 0 in less than 5 minutes (~3 or 4 minutes).

QuentinBisson:
I'm fine with the overall change but I have 2 questions:

1- Do we want to keep a for of 120m for China? (I think we talked about that in a meeting but the info got lost to me)
2- Would it not be best to have 2 alerts (AlertmanagerNotifyNotificationsFailing and AlertmanagerPageNotificationsFailing) instead of having Opsgenie in the name ?

marieroque:
> Opsgenie in the name ?

I'm pro to remove china alerts... and to rename alerts as you suggest.

I did more tests this morning. I know why I'm seeing only 1 log when I set a wrong opsgenie key.
Actually, I see 1 log each 15 minutes, because `group_interval=15m`...

TheoBrigitte:
> I'm pro to remove china alerts... and to rename alerts as you suggest.

I also think removing China alert make sense, because we use a 5mn window and can only alert after 2mn as you already mentioned, and in general special cases should be avoided when possible.

helm/prometheus-rules/templates/alerting-rules/alertmanager.rules.yml
```diff
@@ -11,25 +11,26 @@ spec:
...
-    - alert: AlertManagerNotificationsFailingChina
+        cancel_if_outside_working_hours: "true"
+    - alert: AlertmanagerNotifyNotificationsFailing
       annotations:
         description: '{{`AlertManager {{ $labels.integration }} notifications are failing.`}}'
         opsrecipe: alert-manager-notifications-failing/
-      expr: rate(alertmanager_notifications_failed_total{cluster_id=~"argali|giraffe", integration!~"slack|webhook"}[5m]) > 0
-      for: 120m
+      expr: rate(alertmanager_notifications_failed_total{integration="opsgenie"}[5m]) > 0
+      for: 2m
```
QuentinBisson, line 31: Why not have thé same value for the for field? I would set 10m on both

helm/prometheus-rules/templates/alerting-rules/alertmanager.rules.yml
QuentinBisson: I'm no sure I follow 😅

helm/prometheus-rules/templates/alerting-rules/alertmanager.rules.yml
```diff
@@ -11,25 +11,26 @@ spec:
   groups:
   - name: alertmanager
     rules:
-    - alert: AlertManagerNotificationsFailing
+    - alert: AlertmanagerPageNotificationsFailing
```
QuentinBisson, line 14: I think the expressions are wrong. If intégration is not opsgenie then it is a notify alert and vice versa.

helm/prometheus-rules/templates/alerting-rules/alertmanager.rules.yml
QuentinBisson: But in any case the for should bé thé same for both alerts

helm/prometheus-rules/templates/alerting-rules/alertmanager.rules.yml
marieroque: Well, no it's the opposite.
If opsgenie notification fails, that means we can't request opsgenie, so we can't page, so we notify.

marieroque: Oh yes I see your logic!
And your logic is logic....
I'm so complicated in my head...

QuentinBisson: lool :D

QuentinBisson (approved):

QuentinBisson (approved):

PR #990, opened 2023-12-14
add Kubernetes API server request metrics

TheoBrigitte:
Adding more Kubernetes API server request metrics, in scope of our Hackathon.


TheoBrigitte:
Hackathon is over we had fun. Bye

PR #1001, opened 2024-01-11
Add drift detection between certificates and secrets for certain apps

njuettner:
Towards: https://github.com/giantswarm/giantswarm/issues/29024

As mentioned in the issue, this introduces a new alert on missing
Certificate CR's.

For now I only added `hubble-ui*` and `aws-admission-controller` as a
test.

Here's an example where we would get paged, `anteater` misses the certificate for `aws-admission-controller`, secret is present but `cert-manager` wouldn't renew it:

image

### Checklist

- [ ] Update CHANGELOG.md


Gacko:
So this checks `tls.crt`s in secrets and looks for their corresponding certificate CR just by name, right? What about certificates which set a different secret name?

njuettner:
@Gacko correct, unfortunately it makes it a bit fragile. But almost every app which I checked does have the same naming besides apps which do create certificates different like Kyverno but that was never an issue.

I thought about extending the cert-exporter to add some information like the app name but there simply no information on the resource itself where we could get it.

Gacko:
Normally and if a secret is managed by cert-manager, it sets some annotations. So you can tell if a secret is managed by a certificate by its annotations, IIRC.

njuettner:
@Gacko good input, I created a [PR](https://github.com/giantswarm/cert-exporter/pull/331) for `cert-exporter` to add an additional field for `cert_exporter_secret_...` adding the annotation there which allows me to change the alert slighty.

njuettner:
PR is merged for `cert-exporter` I slightly changed the PromQL so it doesn't rely on the same name for secret and certificate.

image

Gacko (approved):
My PromQL skills aren't that good, but I guess it's fine, especially because you implemented and tested the suggestions. 😬

helm/prometheus-rules/templates/alerting-rules/certificate.management-cluster.rules.yml
```diff
@@ -23,6 +23,18 @@ spec:
         severity: page
         team: rocket
         topic: security
+    - alert: ManagementClusterCertificateIsMissing
+      annotations:
+        description: '{{`Cannot renew Certificate for Secret {{ $labels.namespace }}/{{ $labels.name }} on {{ $labels.cluster_id }} because it is missing.`}}'
+        opsrecipe: managed-app-cert-manager/missing-certificate-for-secret/
```
AndiDog, line 29: This ops-recipe doesn't exist. Is it going to be added @njuettner?

PR #1018, opened 2024-01-31
Compliance metrics

Strigix:
Towards: https://github.com/giantswarm/giantswarm/issues/29011

This PR adds a recording rule for MC cluster compliance metrics, which we can use to create a dashboard to see which clusters are NSA/CIS compliant.


stone-z (approved):
Nice

PR #1022, opened 2024-02-13
Move DNS related alerts to cabbage

T-Kukawka:
during our alerts session we have identified alerts that probably should land on cabbage plate - ping in channel if discussion is needed


T-Kukawka:
@mcharriere updated

QuentinBisson (approved):

kopiczko (approved):

PR #1026, opened 2024-02-23
Add more detailed cluster compliance information

Strigix:
Towards https://github.com/giantswarm/giantswarm/issues/29011
We want to add this recording rule so we can have more detailed information about cluster compliance.


stone-z:
This would include the noisier pod-level checks as well, right? If so, is it possible to avoid sending those to grafana cloud?

PR #1057, opened 2024-03-07
Adjust CAPI rules

njuettner:
Towards: https://github.com/giantswarm/giantswarm/issues/30029

This includes:

- Formatting
- Fixing tests for CAPI
- Changing `MachinePoolReplicasMismatch` and `MachineDeploymentReplicasMismatch` to `MachinePoolIsNotHealthy` and `MachineDeploymentIsNotHealthy` by checking only `Failed` phase
- Changing CAPI rules to opt-out for non business hours until we have confidence rules are working


helm/prometheus-rules/values.yaml
```diff
@@ -2,8 +2,8 @@ name: prometheus-rules
 namespace: monitoring
 serviceType: managed
 project:
-  branch: '[[ .Branch ]]'
-  commit: '[[ .SHA ]]'
+  branch: 'capi-checks'
```
Gacko, line 5: I think you want to revert that once you're done. 😉

njuettner: Good catch, thanks ❤️

QuentinBisson (approved):

PR #1075, opened 2024-03-13
increase PromtailDown 'for' value

QuantumEnigmaa:
Towards https://github.com/giantswarm/giantswarm/issues/30274

This PR increases PromtailDown "for" value to avoid excessive and oftentimes useless paging.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/promtail.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/promtail.rules.yml
@@ -11,7 +11,7 @@ spec:
       rules:
         - alert: PromtailDown
           annotations:
-            description: '{{`Scraping of all promtail pods to check if one failed every 5 minutes.`}}'
+            description: '{{`Scraping of all promtail pods to check if one failed every 30 minutes.`}}'
             opsrecipe: promtail-is-not-running/
           expr: count(up{container="promtail"} == 0) by (cluster_id) > 0
           for: 30m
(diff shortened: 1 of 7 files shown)


helm/prometheus-rules/templates/alerting-rules/promtail.rules.yml
```diff
@@ -11,10 +11,10 @@ spec:
       rules:
         - alert: PromtailDown
           annotations:
-            description: '{{`Scraping of all promtail pods to check if one failed every 5 minutes.`}}'
+            description: '{{`Scraping of all promtail pods to check if one failed every 30 minutes.`}}'
```
QuentinBisson, line 14: Did you check the ksm node metrics? Also, I'm a bit unsure what happens if the container keeps restarting every 2-5 minutes

QuantumEnigmaa: > Did you check the ksm node metrics?

No, what would it be for ?

> Also, I'm a bit unsure what happens if the container keeps restarting every 2-5 minutes

I think we already had this issue with other apps and decided to have 2 different alerts : 1 for when the app is litterally down, and another for when it keeps restarting and/or crashlooping

QuentinBisson: something like that I guess ?

(kube_pod_info * on (cluster_id, pod) group_left() up{container="promtail"}) * on (cluster_id, node) kube_node_status_condition{condition="Ready", status="true"}

QuentinBisson: But let's check with herve :D

QuentinBisson (approved):

PR #1085, opened 2024-03-19
Cilium Handover

T-Kukawka:
Unsure how to change this: https://github.com/giantswarm/prometheus-rules/blob/master/helm/prometheus-rules/templates/recording-rules/service-level.rules.yml

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/cilium.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/cilium.rules.yml
@@ -17,10 +17,10 @@ spec:
       expr: avg(cilium_bpf_map_pressure) by (cluster_id, map_name) * 100 > 80
       for: 15m
       labels:
-        area: kaas
+        area: managedservices
         cancel_if_outside_working_hours: "true"
         severity: page
-        team: phoenix
+        team: cabbage
         topic: cilium
     - alert: CiliumBPFMapFull
       annotations:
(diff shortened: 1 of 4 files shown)


helm/prometheus-rules/templates/recording-rules/service-level.rules.yml
```diff
@@ -41,7 +41,7 @@ spec:
         label_application_giantswarm_io_team: {{ include "providerTeam" . }}
       record: slo_target
 
-      # -- daemonset
+      # -- KAAS daemonset
     - expr: |
         label_replace(
           kube_daemonset_status_desired_number_scheduled{namespace=~"giantswarm|kube-system", daemonset=~"aws-node|aws-cloud-controller-manager|ebs-csi-node|calico-node|cert-exporter|kube-proxy|net-exporter|node-exporter|azure-cloud-controller-manager|azure-cloud-node-manager|azure-scheduled-events|csi-azuredisk-node|cilium"},
```
mcharriere, line 47: `cilium--`

QuentinBisson (approved):

helm/prometheus-rules/templates/alerting-rules/cilium.rules.yml
```diff
@@ -17,10 +17,10 @@ spec:
       expr: avg(cilium_bpf_map_pressure) by (cluster_id, map_name) * 100 > 80
       for: 15m
       labels:
-        area: kaas
+        area: managedservices
```
QuentinBisson, line 20: Is that not platform? I'm confused

T-Kukawka: i based this on other alerts and there were `managedservices` used

QuentinBisson: I do not think you're wrong, we need to sort those labels :)

PR #1090, opened 2024-03-27
Add loki not running alert to avoid high cloud costs

QuentinBisson:

---
Towards: https://github.com/giantswarm/giantswarm/issues/30124 and https://github.com/giantswarm/giantswarm/issues/30381

This PR adds a new alerts that will `transcend all pipelines` to make sure atlas is paged (during business hours) for loki statefulsets not running to prevent high cloud costs due to loki hammering the object storage api if it is crashlooping for any reasons.

We only care about the statefulsets as those are the only components that are accessing S3 but maybe we need the same alert for the loki deployments.

I initially experimented with `increase(loki_s3_request_duration_seconds_count{status_code=~"5.*"}[5m]` but this metric gets reset and rate did not work either so I settled on the statefulset not running alert for 30m

Unit tests are on the way but I wanted eyes first

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/test/tests/providers/global/loki.rules.test.yml
+++ b/test/tests/providers/global/loki.rules.test.yml
@@ -100,16 +100,14 @@ tests:
               opsrecipe: "loki/"
   - interval: 1m
     input_series:
-      - series: 'kube_statefulset_status_replicas{cluster_type="management_cluster", namespace="loki"}'
-        values: "2+0x20 2+0x160"  # 1 unhealthy value after 20 minutes
-      - series: 'kube_statefulset_status_replicas_available{cluster_type="management_cluster", namespace="loki"}'
-        values: "0+0x20 1+0x60 2+0x100"  # 1 pod stop working after 20 minutes then come back up after 60 minutes
+      - series: 'kube_pod_container_status_restarts_total{cluster_type="management_cluster", namespace="loki"}'
+        values: "0+0x20 0+100x20 0+0x120"  # 0 restarts after 20 minutes then we ramp up to 100 restarts in 20 minutes and then back to 0
     alert_rule_test:
-      - alertname: LokiStatefulsetNotRunning
+      - alertname: LokiRestartingTooOften
         eval_time: 15m  # should be OK after 15 minutes
         exp_alerts:
-      - alertname: LokiStatefulsetNotRunning
-        eval_time: 60m  # After 60 minutes, should fire an alert for the t+60 error
+      - alertname: LokiRestartingTooOften
+        eval_time: 85m  # After 85 minutes, should fire an alert for the t+85 error
         exp_alerts:
           - exp_labels:
               all_pipelines: true
@@ -121,8 +119,8 @@ tests:
               team: atlas
               topic: observability
             exp_annotations:
-              description: Loki statefulsets are not running.
+              description: Loki containers are restarting too often.
               opsrecipe: loki/
-      - alertname: LokiStatefulsetNotRunning
-        eval_time: 120m  # After 120m minutes, all should be back to normal
+      - alertname: LokiRestartingTooOften
+        eval_time: 140m  # After 140m minutes, all should be back to normal
         exp_alerts:
--- a/test/tests/providers/global/loki.rules.test.yml
+++ b/test/tests/providers/global/loki.rules.test.yml
@@ -1,6 +1,6 @@
 ---
 rule_files:
-  - loki.all.rules.yml
+  - loki.rules.yml
 
 tests:
   - interval: 1m
@@ -98,3 +98,31 @@ tests:
             exp_annotations:
               description: "Loki pod loki-compactor-676b8c897b-rq298 (namespace loki) sees 1 unhealthy ring members"
               opsrecipe: "loki/"
+  - interval: 1m
+    input_series:
+      - series: 'kube_statefulset_status_replicas{cluster_type="management_cluster", namespace="loki"}'
+        values: "2+0x20 2+0x160"  # 1 unhealthy value after 20 minutes
+      - series: 'kube_statefulset_status_replicas_available{cluster_type="management_cluster", namespace="loki"}'
+        values: "0+0x20 1+0x60 2+0x100"  # 1 pod stop working after 20 minutes then come back up after 60 minutes
+    alert_rule_test:
+      - alertname: LokiStatefulsetNotRunning
+        eval_time: 15m  # should be OK after 15 minutes
+        exp_alerts:
+      - alertname: LokiStatefulsetNotRunning
+        eval_time: 60m  # After 60 minutes, should fire an alert for the t+60 error
+        exp_alerts:
+          - exp_labels:
+              all_pipelines: true
+              area: managedservices
+              cancel_if_outside_working_hours: "true"
+              cluster_type: management_cluster
+              namespace: loki
+              severity: page
+              team: atlas
+              topic: observability
+            exp_annotations:
+              description: Loki statefulsets are not running.
+              opsrecipe: loki/
+      - alertname: LokiStatefulsetNotRunning
+        eval_time: 120m  # After 120m minutes, all should be back to normal
+        exp_alerts:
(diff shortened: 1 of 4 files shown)


QuantumEnigmaa:
How / In which situation can the `kube_statefulset_status_replicas` value be higher than the `kube_statefulset_status_replicas_current` one ?

QuentinBisson:
Fairly easily, `kube_statefulset_status_replicas` is basically `spec.replicas` where `kube_statefulset_status_replicas_current` is the number of running replicas

QuentinBisson:
We could also replace `kube_statefulset_status_replicas_current` with `kube_statefulset_status_replicas_ready` and that should serve the same purpose. The list of statefulset metrics can be found here https://github.com/kubernetes/kube-state-metrics/blob/main/docs/statefulset-metrics.md

QuentinBisson:
The origin of those metrics is that:

```
FIELDS:
   availableReplicas	<integer>
     Total number of available pods (ready for at least minReadySeconds)
     targeted by this statefulset.

   collisionCount	<integer>
     collisionCount is the count of hash collisions for the StatefulSet. The
     StatefulSet controller uses this field as a collision avoidance mechanism
     when it needs to create the name for the newest ControllerRevision.

   conditions	<[]Object>
     Represents the latest available observations of a statefulset's current
     state.

   currentReplicas	<integer>
     currentReplicas is the number of Pods created by the StatefulSet controller
     from the StatefulSet version indicated by currentRevision.

   currentRevision	<string>
     currentRevision, if not empty, indicates the version of the StatefulSet
     used to generate Pods in the sequence [0,currentReplicas).

   observedGeneration	<integer>
     observedGeneration is the most recent generation observed for this
     StatefulSet. It corresponds to the StatefulSet's generation, which is
     updated on mutation by the API Server.

   readyReplicas	<integer>
     readyReplicas is the number of pods created for this StatefulSet with a
     Ready Condition.

   replicas	<integer> -required-
     replicas is the number of Pods created by the StatefulSet controller.

   updateRevision	<string>
     updateRevision, if not empty, indicates the version of the StatefulSet used
     to generate Pods in the sequence [replicas-updatedReplicas,replicas)

   updatedReplicas	<integer>
     updatedReplicas is the number of Pods created by the StatefulSet controller
     from the StatefulSet version indicated by updateRevision.
```

In light of that, maybe we should use `kube_statefulset_status_replicas_available` instead

QuentinBisson:
This would turn the query into 
```promql
kube_statefulset_replicas{cluster_type="management_cluster", namespace="loki"}
  - 
kube_statefulset_status_replicas_available{cluster_type="management_cluster", namespace="loki"}
> 0
```

@giantswarm/team-atlas what do you think?

QuantumEnigmaa:
I feel like `kube_statefulset_status_replicas_available` would suit better here

QuentinBisson:
Moved to kube_statefulset_status_replicas_available

hervenicol:
I'm concerned the pods could be flapping, ie ok/bad/ok/bad at short intervals. Isn't there something like a restart counter we could rely on? Are we sure it never gets to be `available` between 2 restarts and still be a problem?

Also, this alert won't cover the case when it's getting denied access to S3 and hammers network infrastructure with retries (ie proxies). So maybe we should have a separate alert for it.

QuentinBisson:
@hervenicol you're right about flapping and we could use this `kube_pod_container_status_restarts_total` metric for that. Do you think we should make this another alert? I still think we need the alert based on `kube_statefulset_status_replicas_available` because from what I've seen on snail, the pods took some time to be scheduled (i think due to pending state) and this specific alert would haved paged for snail as you can see https://grafana.snail.gaws.gigantic.io/explore?schemaVersion=1&panes=%7B%2287p%22:%7B%22datasource%22:%22PBFA97CFB590B2093%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22kube_statefulset_status_replicas%7Bcluster_type%3D%5C%22management_cluster%5C%22,%20namespace%3D%5C%22loki%5C%22%7D%5Cn%20%20%20%20%20%20%20%20-%20%5Cn%20%20%20%20%20%20%20%20kube_statefulset_status_replicas_current%7Bcluster_type%3D%5C%22management_cluster%5C%22,%20namespace%3D%5C%22loki%5C%22%7D%5Cn%20%20%20%20%20%20%20%20%3E%200%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22PBFA97CFB590B2093%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%22now-30d%22,%22to%22:%22now%22%7D%7D%7D&orgId=1

QuentinBisson:
I agree with you about the network access, I'll try to find one by blocking the CNP but this is not about cost

QuentinBisson:
So both alert makes sense and I'm not sure which one is better than the other.

image


But thinking about the proxy hammering, the issue is that loki was restarting too often so if we have this alert, it should cover the network hammering as we would be alerted about the symptom

hervenicol:
I think the `restarts_total` is easier to read and understand.

QuentinBisson:
It is but it is missing the pending pods but that's okay :)

QuentinBisson:
Changed to pod container restarts

test/tests/providers/global/loki.rules.test.yml
```diff
@@ -98,3 +98,29 @@ tests:
             exp_annotations:
               description: "Loki pod loki-compactor-676b8c897b-rq298 (namespace loki) sees 1 unhealthy ring members"
               opsrecipe: "loki/"
+  - interval: 1m
+    input_series:
+      - series: 'kube_pod_container_status_restarts_total{cluster_type="management_cluster", namespace="loki"}'
+        values: "0+0x20 0+100x20 0+0x120"  # 0 restarts after 20 minutes then we ramp up to 100 restarts in 20 minutes and then back to 0
```
hervenicol, line 104: This is not "100 restarts in 20 minutes" but "100 restarts per minute during 20 minutes".
Quite an unrealistic scenario IMO.
Also, the counter won't go back to 0 in normal situations (although that may also be a test we want to do).

So I'd rather go for:
```suggestion
        values: "0+0x20 0+5x20 100+0x120"  # 0 restarts after 20 minutes then we ramp up to 100 restarts in 20 minutes and then back to 0
```

Or maybe even: 
```
        values: "0+0x20 0+5x20 100+0x120 0+0x120"  # 0 restarts after 20 minutes then we ramp up to 100 restarts in 20 minutes, then stabilize, then test a counter reset to 0
```
(but for this one you may need to add an extra `eval_time` test point)

QuentinBisson: THanks Herve

QuantumEnigmaa (approved):
LGTM but I think we still need another one for S3 access issues (or requests failing in general)

test/tests/providers/global/loki.rules.test.yml
```diff
@@ -98,3 +98,29 @@ tests:
             exp_annotations:
               description: "Loki pod loki-compactor-676b8c897b-rq298 (namespace loki) sees 1 unhealthy ring members"
               opsrecipe: "loki/"
+  - interval: 1m
+    input_series:
+      - series: 'kube_pod_container_status_restarts_total{cluster_type="management_cluster", namespace="loki"}'
+        values: "0+0x20 0+5x20 100+0x120"  # 0 restarts after 20 minutes then we ramp up to 100 restarts in 20 minutes and then back to 0
```
QuentinBisson, line 104:
```suggestion
        values: "0+0x20 0+5x20 100+0x140" # 0 restarts after 20 minutes then we restart 5 times per minute for 20 minutes then we stop restarting for 140 minutes
```

QuentinBisson:
You might be correct @QuantumEnigmaa but this alert (cannot access s3) is not going to work if the container restarts all the time

QuentinBisson:
I still want to experiment with cutting the cnp to test but I'd like to have this out first

PR #1092, opened 2024-03-28
Add recording rule for ingresses using the baseDomain

mcharriere:
Signed-off-by: Matias Charriere <matias@giantswarm.io>

---
Towards: https://github.com/giantswarm/giantswarm/issues/30076

This PR adds recording rule for ingresses using the baseDomain.

It's only valid for Vintage installations, as the value `.Installation.V1.Guest.Kubernetes.IngressController.BaseDomain` is only present there. We need it to get an idea of how many customers are using the base domain in their ingresses.

It should be removed once Vintage is gone.
 
### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
+++ b/helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
@@ -186,6 +186,10 @@ spec:
       record: aggregation:ingress:management_cluster_replicas
     - expr: sum(rate(nginx_ingress_controller_nginx_process_requests_total[5m])) by (cluster_id, cluster_type, customer, installation, pipeline, provider, region)
       record: aggregation:ingress:requests_total
+    {{- with .Values.Installation.V1.Guest.Kubernetes.IngressController.BaseDomain }}
+    - expr: count(count(kube_ingress_path{host=~".*{{ . }}"}) by (ingress,cluster_id)) by (cluster_id)
+      record: aggregation:ingress:base_domain_total
+    {{- end }}
   - name: kubelet.grafana-cloud.recording
     rules:
     - expr: sum(kubelet_running_container_count or kubelet_running_containers{container_state="running"}) by (cluster_id, cluster_type, customer, installation, pipeline, provider, region)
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/values.yaml
```diff
@@ -12,3 +12,10 @@ managementCluster:
 
 mimir:
   enabled: false
+
+Installation:
```
QuentinBisson, line 16: Can we flag this under a better property?

QuentinBisson: You can via giantswarm/shared-configs :)

QuentinBisson: But okay, I won't block you

QuentinBisson (approved):

helm/prometheus-rules/templates/recording-rules/grafana-cloud.rules.yml
```diff
@@ -186,6 +186,10 @@ spec:
       record: aggregation:ingress:management_cluster_replicas
     - expr: sum(rate(nginx_ingress_controller_nginx_process_requests_total[5m])) by (cluster_id, cluster_type, customer, installation, pipeline, provider, region)
       record: aggregation:ingress:requests_total
+    {{- with .Values.Installation.V1.Guest.Kubernetes.IngressController.BaseDomain }}
+    - expr: count(count(kube_ingress_path{host=~".*{{ . }}"}) by (ingress,cluster_id)) by (cluster_id)
```
kopiczko, line 190: Why is this double `count` necessary? `count(kube_ingress_path{host=~".*{{ . }}"}) by (cluster_id)` wont' do the same?

mcharriere: No, note the metric is per path; so ingresses with more than 1 path will have many entries. 

- First count merges the paths for the same ingress in one label.
- Second count actually counts how many ingresses are in the cluster that at least have 1 path with the base domain.

Maybe there's another way to do this, but the query with 1 count will count paths per cluster_id, not ingresses.

PR #1099, opened 2024-04-04
Use namespace of the flux resource in alerts instead of the controller namespace

ljakimczuk:

---

This PR reconfigures Flux part of the monitoring.

### Checklist

- [ ] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
@@ -148,7 +148,7 @@ spec:
           {{`Flux Image Automation Controller on {{ $labels.installation }} seems stuck.`}}
         opsrecipe: flux-image-automation-stuck/
       expr: |
-        sum(irate(workqueue_unfinished_work_seconds{name="imageupdateautomation",cluster_type="management_cluster",exported_namespace="flux-giantswarm"}[15m])) > 0
+        sum(irate(workqueue_unfinished_work_seconds{name="imageupdateautomation",cluster_type="management_cluster",namespace=~"flux-giantswarm|flux-system"}[15m])) > 0
       for: 30m
       labels:
         area: empowerment
--- a/helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
@@ -30,7 +30,7 @@ spec:
         description: |-
           {{`Flux HelmRelease {{ $labels.name }} in ns {{ $labels.namespace }} on {{ $labels.installation }}/{{ $labels.cluster_id }} is stuck in Failed state.`}}
         opsrecipe: fluxcd-failing-helmrelease/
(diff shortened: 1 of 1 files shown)


helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
```diff
@@ -60,7 +60,7 @@ spec:
         description: |-
           {{`Flux Kustomization {{ $labels.name }} in ns {{ $labels.namespace }} on {{ $labels.installation }}/{{ $labels.cluster_id }} is stuck in Failed state.`}}
         opsrecipe: fluxcd-failing-kustomization/
-      expr: gotk_reconcile_condition{app="flux-giantswarm-monitoring", type="Ready", status="False", kind="Kustomization", cluster_type="management_cluster", namespace=~".*giantswarm.*"} > 0
+      expr: gotk_reconcile_condition{type="Ready", status="False", kind="Kustomization", cluster_type="management_cluster", exported_namespace=~".*giantswarm.*"} > 0
```
QuentinBisson, line 63: Why is this a regex while it is not below?

helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
```diff
@@ -148,7 +148,7 @@ spec:
           {{`Flux Image Automation Controller on {{ $labels.installation }} seems stuck.`}}
         opsrecipe: flux-image-automation-stuck/
       expr: |
-        sum(irate(workqueue_unfinished_work_seconds{name="imageupdateautomation",cluster_type="management_cluster",namespace="flux-system"}[15m])) > 0
+        sum(irate(workqueue_unfinished_work_seconds{name="imageupdateautomation",cluster_type="management_cluster",namespace=~"flux-giantswarm|flux-system"}[15m])) > 0
```
QuentinBisson, line 151: Should you use namespace or exported namespace? Now I'm confused

QuentinBisson:
Also, could you maybe use this opportunity to add tests?

helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
uvegla: Flux exposed ones already have a namespace label. These are exposed by kubebuilder and they dont have it so the name is not mangled.

QuentinBisson: I'm not sure I follow

helm/prometheus-rules/templates/alerting-rules/flux.rules.yml
ljakimczuk: The reason is to keep the original intention behind these metrics, but to adjust them slightly. So if the metric was designed previously to not use a regex, it is not using it now, after adjustment. If it has been designed to use regex, it still uses regex.

QuentinBisson (approved):

PR #1139, opened 2024-04-24
Update LokiRingUnhealthy query to avoid false positive

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/30469

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/loki.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/loki.rules.yml
@@ -73,7 +73,7 @@ spec:
         description: '{{`Loki pod {{ $labels.pod }} (namespace {{ $labels.namespace }}) sees {{ $value }} unhealthy ring members`}}'
         opsrecipe: loki/
       expr: |
-        sum by (app, cluster_id, container, customer, installation, provider, pipeline, name, namespace, organization, pod) (cortex_ring_members{state="Unhealthy"}) > 0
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/alerting-rules/loki.rules.yml
```diff
@@ -73,7 +73,8 @@ spec:
         description: '{{`Loki pod {{ $labels.pod }} (namespace {{ $labels.namespace }}) sees {{ $value }} unhealthy ring members`}}'
         opsrecipe: loki/
       expr: |
-        sum (min_over_time(cortex_ring_members{state="Unhealthy"}[30m])) by (app, cluster_id, container, customer, installation, provider, pipeline, name, namespace, organization, pod) > 0
```
QuentinBisson, line 76: Do WE want this to page only when app=Loki?

marieroque: You mean with Mimir, the query can match mimir pods ?

QuentinBisson: I think so yes

hervenicol:
How did you check the new query solves the issue?
I was thinking UTs could be updated to fail with the previous query, and then query would be improved to pass the tests.

marieroque:
> inking UTs could be updated to fail with the previous query, and then query would be improved to pass the te

I tested the query by comparing it on a previous incident on potato.
UT didn't need to be updated as I kept 30 minutes of duration.

marieroque:
I tested previous alert (min_over_time) and new alert (min with for) by adding a new serie in TU.

I could see the previous alert is failing the TUs whereas the new alert passes the TUs.

PR #1153, opened 2024-04-30
Add inhibitions expressions for CAPI clusters

weseven:
Towards: https://github.com/giantswarm/roadmap/issues/3338
This PR adds the expressions to have inhibitions working on capi clusters.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alerting-rules/inhibit.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/inhibit.management-cluster.rules.yml
@@ -80,7 +80,7 @@ spec:
     - alert: InhibitionClusterNodePoolsNotReady
       annotations:
         description: '{{`Cluster {{ $labels.cluster_id }} node pools are not ready. Either they have been scaled down to 0 or they are not up yet.`}}'
-      expr: cluster_operator_node_pool_desired_workers == 0 and cluster_operator_node_pool_ready_workers == 0 or capi_machinepool_status_condition{type="Ready", status="False"} == 1
+      expr: (cluster_operator_node_pool_desired_workers == 0 and cluster_operator_node_pool_ready_workers == 0) or capi_machinepool_status_condition{type="Ready", status="False"} == 1
       labels:
         area: kaas
         cluster_with_notready_nodepools: "true"
--- a/helm/prometheus-rules/templates/alerting-rules/inhibit.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/alerting-rules/inhibit.management-cluster.rules.yml
@@ -17,7 +17,7 @@ spec:
     - alert: InhibitionClusterStatusCreating
       annotations:
         description: '{{`Cluster {{ $labels.cluster_id }} has status condition ''Creating''.`}}'
-      expr: label_replace(max_over_time(statusresource_cluster_status{status="Creating"}[30m]), "cluster_id", "$1", "exported_cluster_id", "(.+)") == 1 or max_over_time(cluster_operator_cluster_status{status="Creating"}[30m]) == 1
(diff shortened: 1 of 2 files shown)


QuentinBisson (approved):

njuettner (approved):

helm/prometheus-rules/templates/alerting-rules/inhibit.management-cluster.rules.yml
```diff
@@ -80,7 +80,7 @@ spec:
     - alert: InhibitionClusterNodePoolsNotReady
       annotations:
         description: '{{`Cluster {{ $labels.cluster_id }} node pools are not ready. Either they have been scaled down to 0 or they are not up yet.`}}'
-      expr: cluster_operator_node_pool_desired_workers == 0 and cluster_operator_node_pool_ready_workers == 0
+      expr: cluster_operator_node_pool_desired_workers == 0 and cluster_operator_node_pool_ready_workers == 0 or capi_machinepool_status_condition{type="Ready", status="False"} == 1
```
AverageMarcus, line 83: Does this now need to be in parenthesis? I'm not sure what the order of operations is for promql here.

weseven: Good catch; looking at [prometheus docs, or seems to have the lowest priority](https://prometheus.io/docs/prometheus/latest/querying/operators/#binary-operator-precedence), but still, with parenthesis it's clearer for everyone :)
Thanks, added them in the last commit!

helm/prometheus-rules/templates/alerting-rules/inhibit.management-cluster.rules.yml
```diff
@@ -80,7 +80,7 @@ spec:
     - alert: InhibitionClusterNodePoolsNotReady
       annotations:
         description: '{{`Cluster {{ $labels.cluster_id }} node pools are not ready. Either they have been scaled down to 0 or they are not up yet.`}}'
-      expr: cluster_operator_node_pool_desired_workers == 0 and cluster_operator_node_pool_ready_workers == 0
+      expr: (cluster_operator_node_pool_desired_workers == 0 and cluster_operator_node_pool_ready_workers == 0) or capi_machinepool_status_condition{type="Ready", status="False"} == 1
```
AverageMarcus, line 83: Thinking about it some more, what happens if there are multiple node pools? (e.g. one per AZ as is recommend) This wouldn't equal to `1` would it?

weseven: Correct, this would equal to 1 even if only one nodePool among N is not ready.
Is that not the desired behaviour? I see the `cancel_if_cluster_with_notready_nodepools` is used in DNS errorrate, APIserver latency and network error rate alerts, so I assumed even only one nodepool not ready would affect those.

weseven: I think it works the same way as in vintage, there's a `node_pool_id` label on the `cluster_operator_node_pool_desired_workers` and `cluster_operator_node_pool_ready_workers` metrics.
So for example it is currently triggering for deu01, which has a mix of active nodepools with nodes and nodepools scaled to 0.
It does sound odd though the more I think about it... @QuentinBisson, do you have any historical context on this? (Just because git blame points to your commit on these lines :sweat_smile: )

QuentinBisson: I have no context over this, this was moved from another repo 3 years ago, What you need to be sure about is that the inhibitoin and alert can relate to a nodepool label but I'm afraid we do not have it in alerts that use this inhibition?

weseven: (maybe it could be useful for the cluster downscaler?)

QuentinBisson: Sure you can add and remove :D

PR #1193, opened 2024-05-27
Fix incorrect ksm alert fixing on mimir

QuentinBisson:

---
This PR fixes the ongoing ksm alerts on grizzly, caused by some leftover slos (data being kept for 30 days, the initial query was counting them so firing for the wrong reasons).

Adding the regex of __name__ ensures we only count the ksm generated alerts in the query.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix node load alerts for CAPI clusters.
 - Remove trailing spaces in rules.
+- Fix `KubeStateMetricsNotRetrievingMetrics` on mimir.
 
 ## [3.14.2] - 2024-05-16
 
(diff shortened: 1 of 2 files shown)


QuantumEnigmaa (approved):

hervenicol:
Can't we limit the query to more recent data? Because if it takes old data into account, it won't work as expected (ie alerting when there's no recent data from KSM)?

QuentinBisson:
It does not take old data into account as it's using a count

hervenicol:
oh, it's SLO-generated data, but not old data!
image

QuentinBisson:
Yes that's what I explained by:

```
caused by some leftover slos (data being kept for 30 days, the initial query was counting them so firing for the wrong reasons).
```

hervenicol:
Yes, but when I read "data being kept for 30 days" I understand "old data" :sweat_smile: 
So the new query looks like this, which looks good :+1:  
image

hervenicol (approved):

QuentinBisson:
Good to know how I can improve :)

PR #1218, opened 2024-06-06
Review turtles alerts labels

weseven:
Towards: https://github.com/giantswarm/giantswarm/issues/30934
This PR removes the labels app, node and role where possible/where it makes sense from turtles alerts.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/shared/alerting-rules/node.management_cluster.rules.yml
+++ b/helm/prometheus-rules/templates/shared/alerting-rules/node.management_cluster.rules.yml
@@ -26,9 +26,9 @@ spec:
         description: '{{`Node {{ $labels.node }} status is flapping under load.`}}'
       expr: |
         (
-          sum(node_load15{cluster_type="management_cluster", app!="vault", role!="bastion"})
+          sum(node_load15{cluster_type="management_cluster", service="node-exporter"})
             by (cluster_id, installation, node, pipeline, provider)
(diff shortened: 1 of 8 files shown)


njuettner (approved):

helm/prometheus-rules/templates/shared/alerting-rules/node.management_cluster.rules.yml
```diff
@@ -26,9 +26,9 @@ spec:
         description: '{{`Node {{ $labels.node }} status is flapping under load.`}}'
       expr: |
         (
-          sum(node_load15{cluster_type="management_cluster", app!="vault", role!="bastion"})
+          sum(node_load15{cluster_type="management_cluster", service="node-exporter"})
```
QuentinBisson, line 29: Are you sure you want to be alerted for bastions as well?

weseven: I've checked on anteater, for the bastions the service label does not have a value; could it be different for other MCs? (if so, which are the minimal test cases? Ideally I want to avoid checking every single one for label disparities :see_no_evil: )

QuentinBisson: I think it's fine :)

QuentinBisson (approved):

PR #1224, opened 2024-06-06
Use ready replicas for KyvernoWebhookHasNoAvailableReplicas

fhielpos:
From: https://github.com/giantswarm/giantswarm/issues/31003

Uses ready replicas instead of just the current status replicas for the KyvernoWebhookHasNoAvailableReplicas alert. This ensure the alert fires even when pods are Pending.


Strigix:
Do we still wanna keep it at 15 minutes??
What do you think?

stone-z (approved):
Plz link the issue to the PR but LGTM

PR #1246, opened 2024-06-14
remove duplicate slo-target on AWS

hervenicol:
Towards incident: [#inc-2024-06-14-all-clusters-prometheusrulesfailures](https://gigantic.slack.com/archives/C078HF7820Z)

We now have all rules deployed on all installations.
For slo-target recording rules, we had some metrics that were generic, and some that were aws-specific.
It seems for `service="kube-controller-manager"` both were valid, thus generating duplicate metrics.

With this PR, we stop using the aws-specific metrics and keep only the generic one.

### Checklist

- [x] Update CHANGELOG.md


helm/prometheus-rules/templates/platform/atlas/recording-rules/service-level.rules.yml
```diff
@@ -317,21 +317,21 @@ spec:
 
     # core k8s components aws API requests
     # record number of requests.
-    - expr: label_replace(sum(cloudprovider_aws_api_request_duration_seconds_count{job=~"kube-controller-manager|kube-scheduler"}) by (cluster_id, cluster_type, customer, installation, pipeline, provider, region, job), "service", "$1", "job", "(.*)")
+    - expr: label_replace(sum(cloudprovider_aws_api_request_duration_seconds_count{job="kube-scheduler"}) by (cluster_id, cluster_type, customer, installation, pipeline, provider, region, job), "service", "$1", "job", "(.*)")
```
QuentinBisson, line 320: I think maybe thé service label should bé différent liké kube-controller-manager-aws-cloud-api? And I Guess azure for the other one? Also, I think it's obvious the cloud api metric does not exist for the scheduler or it should go away

hervenicol: There's also the cluster_id in the alert, so I'm not sure we need a provider-specific metrics name.

QuentinBisson: You right because it's unlikely we will have a cluster with 2 providers so only ones of both metrics I'll be active for a cluster

QuentinBisson: Then we probably should copy the name from hère: https://github.com/giantswarm/sloth-rules/blob/main/areas/kaas/teams/phoenix/cloudapis/slos/availability-aws.yaml

QuentinBisson: Or maybe we Can get rid of this one hère if we have the sloth ones. @giantswarm/team-turtles or @giantswarm/team-phoenix  any thoughts ? Is the New sloth ones silences?

hervenicol: But if we change the name of the slo_targets, it means we have to change all the rules that depend on it, right? And stop having some generic ones?
We're talking about recording rules like `slo_threshold_high` and `slo_threshold_low` here, quite generic ones!

TheoBrigitte (approved):
LGTM, when tested

helm/prometheus-rules/templates/platform/atlas/recording-rules/service-level.rules.yml
QuentinBisson: I'm not talking about changing thé name of the metrics but if the service label :)

QuentinBisson: So

```
label_replace(xxx, "service", "$1", "job", "(.*)")
```

Would become

```
label_replace(xxx, "service", "$1-cloudapis", "job", "(.*)")
```

hervenicol:
Tested on `gaia`, it fixed the failing prometheus rule.

PR #1285, opened 2024-07-05
Replace grafana-agent with alloy

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/3521

Replace Grafana Agent with Alloy to import PrometheusRules into Mimir.

In this case Alloy is a drop-in replacement for Grafana Agent. So config stays exactly the same.


helm/prometheus-rules/templates/grafana-agent-rules-configmap.yaml
```diff
@@ -1,41 +1,13 @@
 {{- if .Values.mimir.enabled }}
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: alloy-rules-config
+  namespace: giantswarm
 data:
   values: |
-    networkPolicy:
-      cilium:
```
QuentinBisson, line 6: Why did you remove this policy configuration ?

TheoBrigitte: Because the networkpolicy is directly within the helm chart of Alloy.

QuentinBisson: We need to push it  upstream then because @marieroque will not like this :D

helm/prometheus-rules/templates/alloy-rules.yaml
```diff
@@ -5,13 +5,13 @@ metadata:
   labels:
     app-operator.giantswarm.io/version: 0.0.0
     {{- include "labels.common" . | nindent 4 }}
-  name: grafana-agent-rules
+  name: alloy-rules
   namespace: giantswarm
 spec:
-  catalog: giantswarm
+  catalog: giantswarm-playground
   config:
```
QuentinBisson, line 12: Can you move alloy out of playground?

TheoBrigitte: Which catalog do we want it into ?

QuentinBisson: Either giantswarm or default I think

TheoBrigitte: done

QuentinBisson: Thanks :)

QuentinBisson:
I'm fine with this but i'll let @marieroque take a look

helm/prometheus-rules/templates/grafana-agent-rules-configmap.yaml
marieroque: Maybe we need to add the CNP:
https://github.com/giantswarm/grafana-agent-app/blob/main/helm/grafana-agent/templates/cilium-networkpolicy.yaml
as manifest in that repo ?

marieroque: @TheoBrigitte did your tests work in term of communication ?
Did Hubble UI show no dropped request ?

TheoBrigitte: I did not check Hubble UI, my test was successful (I deleted all the rules from Mimir, and Alloy re-imported them all)
Also there is a CNP already added in Alloy chart > https://github.com/giantswarm/alloy-app/blob/main/helm/alloy/templates/cilium-networkpolicy.yaml

TheoBrigitte: Just checked and there's no error in Hubble UI :heavy_check_mark:

QuentinBisson (approved):

QuentinBisson:
Before merging, how about we rename it to something that does not include the project name in it

marieroque (approved):

TheoBrigitte:
> Before merging, how about we rename it to something that does not include the project name in it like mimir-rules-loader or smth? That way if we change to something else in thé future the component name does not change?

There might be a change that we will use this Alloy to also load rules into Loki, so I will keep this name for now and we can always rename this later on if there's a need for it.

QuentinBisson:
Sure sure and we can rename later. I would like if we do not do the same mistake with the bundle later on when we introduce alloy 😅

QuentinBisson:
I would hope that you remember this component is alloy marie 😅

PR #1286, opened 2024-07-08
disabel aggregation:kyverno_policy_deployment_status_team recording rule

QuantumEnigmaa:
Towards: https://github.com/giantswarm/giantswarm/issues/24418

This PR disables the `aggregation:kyverno_policy_deployment_status_team` rule to save on Grafana Cloud costs. After a quick calculation, this should help save ~$300 monthly.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Disabled `aggregation:kyverno_policy_deployment_status_team` recording rule to save on Grafana Cloud cost.
+
 ## [4.5.0] - 2024-07-03
 
 ### Changed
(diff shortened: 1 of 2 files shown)


TheoBrigitte (approved):
Why not just deleting it instead of commenting ? We can always find this back in git history,

QuantumEnigmaa:
I just personally find it easier for reverting this change (which will eventually happen at some point) to have it commented rather than having to search in the git history later :shrug:

hervenicol:
Why will we eventually revert this change?
Do we need this data, or not?

QuantumEnigmaa:
See the following slack thread : https://gigantic.slack.com/archives/C02FL4EAADD/p1720429374452249

hervenicol:
I don't really like that we have a recording rule that we enable/disable like that to save costs.
I'd rather have a sustainable solution that we can keep running all the time at a lower cost.

I tent to think `policy_report_result` has too high cardinality, and should not be generated as metrics.
Maybe it should be some kind of event stored in loki.

hervenicol (approved):
Fine for me for now, even though I'd love to keep an eye on `policy_report_results` metrics because they have a really high cardinality.

QuantumEnigmaa:
I understand your concerns and I also think a more persistent solution in the long run. But yeah, for now, just disabling this rule could allow us to really save on costs.

PR #1297, opened 2024-07-14
add support for loki rules to management clusters in alloy config

hervenicol:
Towards https://github.com/giantswarm/roadmap/issues/3178

This PR adds:
- support for loading loki rules into Loki's management-cluster tenant.
- a grafana datasource so we can see them on grafana

prometheusRules for loki should have a label `application.giantswarm.io/prometheus-rule-kind: loki`.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/alloy-rules-configmap.yaml
+++ b/helm/prometheus-rules/templates/alloy-rules-configmap.yaml
@@ -45,6 +45,24 @@ data:
             mimir.rules.kubernetes "local" {
               address = "http://mimir-ruler.mimir.svc:8080/"
               tenant_id = "anonymous"
+              rule_selector {
+                  match_expression {
+                    key = "application.giantswarm.io/prometheus-rule-kind"
+                    operator = "NotIn"
+                    values = ["loki"]
+                  }
+              }
+            }
+            loki.rules.kubernetes "local" {
+              address = "http://loki-backend.loki.svc:3100/"
+              tenant_id = "{{ .Values.managementCluster.name }}"
+              rule_selector {
+                  match_expression {
+                    key = "application.giantswarm.io/prometheus-rule-kind"
+                    operator = "In"
+                    values = ["loki"]
+                  }
+              }
             }
       controller:
         type: "deployment"
--- a/helm/prometheus-rules/templates/loki-ruler-datasource-configmap.yaml
+++ b/helm/prometheus-rules/templates/loki-ruler-datasource-configmap.yaml
@@ -1,0 +1,23 @@
+{{- if .Values.mimir.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: loki-ruler-datasource-{{ .Values.managementCluster.name }}
+  namespace: monitoring
+  labels:
+    app.giantswarm.io/kind: datasource
+data:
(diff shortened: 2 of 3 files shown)


helm/prometheus-rules/templates/loki-ruler-datasource-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+  loki-ruler-{{ .Values.managementCluster.name }}.yaml: |
+    apiVersion: 1
+    datasources:
+    - access: proxy
+      editable: false
+      jsonData:
+        manageAlerts: true
+        httpHeaderName1: X-Scope-OrgID
+      secureJsonData:
+        httpHeaderValue1: {{ .Values.managementCluster.name }}
```
QuentinBisson, line 19: Any reason why we add a tenant info here? Mimir multi-tenancy is not enabled IIRC

QuentinBisson: Oh that's in loki my bad

helm/prometheus-rules/templates/loki-ruler-datasource-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+      name: Loki-ruler-{{ .Values.managementCluster.name }}
+      type: loki
+      url: http://loki-backend.loki.svc:3100
```
QuentinBisson, line 22: What's the difference between this datasource and the loki one?

helm/prometheus-rules/templates/alloy-rules-configmap.yaml
```diff
@@ -45,6 +45,24 @@ data:
             mimir.rules.kubernetes "local" {
               address = "http://mimir-ruler.mimir.svc:8080/"
               tenant_id = "anonymous"
+              rule_selector {
```
QuentinBisson, line 48: Maybe we should have alloy in a different namespace than mimir in another PR?

QuentinBisson (approved):

helm/prometheus-rules/templates/loki-ruler-datasource-configmap.yaml
hervenicol: This one queries the backend directly, which is responsible for the rules. And with a tenant ID.
The reason for this is that when we go through the gateway, the MTP adds a list of tenants, and the ruler does not work with that.
So, as a first step, the rules will only work on the MC logs. And so does this extra datasource.

This datasource won't work for logs queries, it only works for listing rules.

I know it's not an ideal solution, but at least we can have a MVP to start with, then we can look for improvements.

helm/prometheus-rules/templates/alloy-rules-configmap.yaml
hervenicol: I agree, alloy probably belongs to `monitoring` namespace.

helm/prometheus-rules/templates/loki-ruler-datasource-configmap.yaml
QuentinBisson: Sure that makes sense, thanks for explaining

PR #1305, opened 2024-07-17
add LokiNeedsToBeScaledUp alert

QuantumEnigmaa:
Towards: https://github.com/giantswarm/roadmap/issues/3534

This PR adds a `LokiNeedsToBeScaledUp` alert to notify oncalers that loki is going to be scaled up by the hpa. This means that if an oncaller gets paged by this, he/she will need to make sure that no previous user-values or any other change has been made to the loki workload that would compere with the upscaling. 

In particular, if some resources are lowered down by a user-values compared to what's defined in the shared-configs or in the installation's specific config repo,  the oncaller will have to remove that configmap. 

This will be detailed in the loki ops recipe once I'll have updated it.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
+++ b/helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
@@ -83,15 +83,15 @@ spec:
         severity: page
         team: atlas
         topic: observability
-    - alert: LokiNeedsToBeScaledUp
+    - alert: LokiHpaReachedMaxReplicas
       annotations:
-        description: '{{`Loki component {{ $labels.horizontalpodautoscaler }} needs to be scaled up.`}}'
+        description: '{{`Loki component {{ $labels.horizontalpodautoscaler }} has reached its maxReplicas number but still needs to be scaled up.`}}'
         opsrecipe: loki/
       expr: |
         sum by (cluster_id, installation, namespace, pipeline, provider, horizontalpodautoscaler) (kube_horizontalpodautoscaler_status_desired_replicas{namespace="loki", horizontalpodautoscaler=~"loki-backend|loki-write|loki-read"})
           != 
         sum by (cluster_id, installation, namespace, pipeline, provider, horizontalpodautoscaler) (kube_horizontalpodautoscaler_status_current_replicas{namespace="loki", horizontalpodautoscaler=~"loki-backend|loki-write|loki-read"})
-      for: 10m
+      for: 4h
       labels:
         area: platform
         cancel_if_cluster_control_plane_unhealthy: "true"
--- a/helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
+++ b/helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
@@ -83,3 +83,22 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: LokiNeedsToBeScaledUp
+      annotations:
+        description: '{{`Loki component {{ $labels.horizontalpodautoscaler }} needs to be scaled up.`}}'
+        opsrecipe: loki/
+      expr: |
+        sum by (cluster_id, installation, namespace, pipeline, provider, horizontalpodautoscaler) (kube_horizontalpodautoscaler_status_desired_replicas{namespace="loki", horizontalpodautoscaler=~"loki-backend|loki-write|loki-read"})
+          != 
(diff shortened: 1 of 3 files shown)


hervenicol:
> This PR adds a `LokiNeedsToBeScaledUp` alert to notify oncalers that loki is going to be scaled up by the hpa.

I think it rather alerts that the HPA would like to scale up higher but can't.
If the HPA is going to act (and fix the alert) automatically, we don't want to get paged. So 10min looks short to me.

Also, on some clusters that are quite resource-tight, we may have voluntarily set a low maxReplicas as per the opsrecipe. So that alert would be constantly firing. What is the remediation for these cases? Set an infinite silence?

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
```diff
@@ -83,3 +83,22 @@ spec:
...
         topic: observability
+    - alert: LokiNeedsToBeScaledUp
+      annotations:
+        description: '{{`Loki component {{ $labels.horizontalpodautoscaler }} needs to be scaled up.`}}'
+        opsrecipe: loki/
+      expr: |
+        sum by (cluster_id, installation, namespace, pipeline, provider, horizontalpodautoscaler) (kube_horizontalpodautoscaler_status_desired_replicas{namespace="loki", horizontalpodautoscaler=~"loki-backend|loki-write|loki-read"})
+          != 
+        sum by (cluster_id, installation, namespace, pipeline, provider, horizontalpodautoscaler) (kube_horizontalpodautoscaler_status_current_replicas{namespace="loki", horizontalpodautoscaler=~"loki-backend|loki-write|loki-read"})
+      for: 10m
```
QuentinBisson, line 94: As @hervenicol said I think 10 minutes is too short. I would probably make this 1h.

Also  I would think the alert should be named something like LokiHpaNotWorking or something like LokiHpaReplicasMismatch ?  I took it from here https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/3dfa72d1d1ab31a686b1f52ec28bbf77c972bd23/alerts/apps_alerts.libsonnet#L284

QuentinBisson: I would also like if we were not paged if we reached the max number of replicas which is probably something we can get from the HPA?

QuantumEnigmaa: I am concerned by a specific scenario : someone has dramatically reduced resources requests for a loki component following a `LokiNeedsToBeScaledDown` alert but at some point, actual usage does increase which leads to the hpa scaling up the number of replicas. This would lead to having a lot of replicas with very low resources requests whereas we could have had less replcias with higher resources requests which is better right ?

Concerning the time period, I'm totally ok to increase it, 10m is indeed quite short. 

I'm also ok with renaming the alert.

Concerning your identical concern in the case of the component reaching the maxReplicas number I think it relates to my first paragraph right ? If the hpa still wants to scale up the component while being already at maxReplicas, this can mean we're in the scenario I described and that we should act, unless hpa can't recommend a replica number higer than it maxReplicas.

hervenicol: > I am concerned by a specific scenario : someone has dramatically reduced resources requests for a loki component following a `LokiNeedsToBeScaledDown` alert but at some point, actual usage does increase which leads to the hpa scaling up the number of replicas. This would lead to having a lot of replicas with very low resources requests whereas we could have had less replcias with higher resources requests which is better right ?

In my opinion, this was exactly the reason why we wanted a "scale up" alert.
But I think your query does not match the use case.

> I'm also ok with renaming the alert.

Well, the proposed names are to match the current alert. But I'm not even sure we want the current alert. :sweat_smile: 

> If the hpa still wants to scale up the component while being already at maxReplicas, this can mean we're in the scenario I described and that we should act, unless hpa can't recommend a replica number higer than it maxReplicas.

Oh, so you want an alert that fires when the component reaches MaxReplicas but still requires more replicas?

I'm rather fine with this, except:

> on some clusters that are quite resource-tight, we may have voluntarily set a low maxReplicas as per the opsrecipe. So that alert would be constantly firing. What is the remediation for these cases? Set an infinite silence?

QuantumEnigmaa: Let's talk about all of this after a standup or during a dedicated meeting :sweat_smile:

QuantumEnigmaa: Following our discussions, I renamed the alert and increased the time period for triggering it to 4h.

QuantumEnigmaa:
@hervenicol Are you ok to merge this one and the other now that the ops recipe has been updated ?

hervenicol (approved):

PR #1317, opened 2024-08-05
remove cluster_id label from 2 loki alerting rules

QuantumEnigmaa:
This PR removes the `cluster_id` label from both `LokiHpaReachedMaxReplicas` and `LokiNeedsToBeScaledDown` alerting rules as this leads to situations where the total amount of resources requested by each loki component is calculated for each cluster which means the results are faulty.

For example in `gazelle` : 

- With the `cluster_id` label : 
image

- Without it : 
image

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Removed `cluster_id` label from `LokiHpaReachedMaxReplicas` and `LokiNeedsToBeScaledDown` rules.
+
 ## [4.9.0] - 2024-08-01
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Removed `cluster_id` label from `LokiHpaReachedMaxReplicas` and `LokiNeedsToBeScaledDown` rules.
+- Restricted range of `LokiHpaReachedMaxReplicas` and `LokiNeedsToBeScaledDown` rules to management clusters.
 
 ## [4.9.0] - 2024-08-01
 
--- a/helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
+++ b/helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
@@ -88,14 +88,14 @@ spec:
         description: '{{`Loki component {{ $labels.labelpod }} is consuming very few resources and needs to be scaled down.`}}'
         opsrecipe: loki/
       expr: |-
-        sum by (cluster_id, installation, namespace, pipeline, provider, labelpod) (label_replace(container_memory_working_set_bytes{container="loki", namespace="loki"}, "labelpod", "$1", "pod", "(loki-[[:alnum:]]*)-.*")) 
+        sum by (installation, namespace, pipeline, provider, labelpod) (label_replace(container_memory_working_set_bytes{container="loki", namespace="loki"}, "labelpod", "$1", "pod", "(loki-[[:alnum:]]*)-.*")) 
           / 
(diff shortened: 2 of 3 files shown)


TheoBrigitte (approved):

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
```diff
@@ -88,14 +88,14 @@ spec:
         description: '{{`Loki component {{ $labels.labelpod }} is consuming very few resources and needs to be scaled down.`}}'
         opsrecipe: loki/
       expr: |-
-        sum by (cluster_id, installation, namespace, pipeline, provider, labelpod) (label_replace(container_memory_working_set_bytes{container="loki", namespace="loki"}, "labelpod", "$1", "pod", "(loki-[[:alnum:]]*)-.*"))
```
QuentinBisson, line 91: We need this label for alerts though

hervenicol:
Could you show which query shows different clusters?
Maybe this would end up showing a buggy query.

Otherwise, if we are sure all is ok when we understand where these WC cluster_id come from, you can label_replace the cluster_id with the MC name.
But I'd like to make sure the query is good before doing that, and we're not actually retrieving WC data.

QuantumEnigmaa:
So the query showing different results is the one currently used for the alert : 
```
        sum by (cluster_id, installation, namespace, pipeline, provider, labelpod) (label_replace(container_memory_working_set_bytes{container="loki", namespace="loki"}, "labelpod", "$1", "pod", "(loki-[[:alnum:]]*)-.*")) 
          / 
        sum by(cluster_id, installation, namespace, pipeline, provider, labelpod) (label_replace(kube_pod_container_resource_requests{container="loki", namespace="loki", unit="byte"}, "labelpod", "$1", "pod", "(loki-[[:alnum:]]*)-.*")) 
          <= 0.30 
        and
        sum(label_replace(rate(container_cpu_usage_seconds_total{container="loki", namespace="loki"}[5m]), "labelpod", "$1", "pod", "(loki-[[:alnum:]]*)-.*")) by (cluster_id, installation, namespace, pipeline, provider, labelpod) 
          / 
        sum by(cluster_id, installation, namespace, pipeline, provider, labelpod) (label_replace(kube_pod_container_resource_requests{container="loki", namespace="loki", unit="core"}, "labelpod", "$1", "pod", "(loki-[[:alnum:]]*)-.*")) 
          <= 0.30
```

But for some reason this is only happening on `gazelle`. On all other installations, I'm only getting data from the MC as expected and removing the `cluster_id` label there doesn't change the graphs and values displayed.

hervenicol:
> But for some reason this is only happening on `gazelle`.

Maybe because gazelle has loki deployed on some WCs?

hervenicol:
So,  I suggest you limit your rule to MCs by adding `cluster_type="management_cluster"`

For instance:
```
kube_pod_container_resource_requests{container="loki", namespace="loki", unit="byte"}
```
becomes:
```
kube_pod_container_resource_requests{container="loki", namespace="loki", unit="byte", cluster_type="management_cluster"}
```

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
hervenicol: BTW, great catch from `pint`! :tada:

QuantumEnigmaa:
> Maybe because gazelle has loki deployed on some WCs?

Yeah you're absolutely right, I didn't thought about it but after checking, the `cicdprod` WC on `gazelle` is indeed running loki.

> I suggest you limit your rule to MCs by adding cluster_type="management_cluster"

I think that's the best thing to do indeed :)

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Removed `cluster_id` label from `LokiHpaReachedMaxReplicas` and `LokiNeedsToBeScaledDown` rules.
```
QuentinBisson, line 12: You probably want to change thé entry

QuentinBisson: Otherwise LGTM. Great Idea @hervenicol

QuantumEnigmaa: Done.

PR #1324, opened 2024-08-20
Fix prometheus agent failing alert for multi-provider mcs

QuentinBisson:

---
Towards: https://github.com/giantswarm/...

This PR fixes issues with prometheus-agent-failing alert on multi-provider MCs because the provider label is not the same between cluster-api-monitoring metrics and the WC metrics.

@giantswarm/team-atlas if you find a better solution, we can definitely revert it :)

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Upgrade Alloy to 0.3.1
 
+### Fixed
+
+- Fix Prometheus Agent Failing alert for Multi-provider MCs.
+
 ## [4.9.1] - 2024-08-06
 
 ### Changed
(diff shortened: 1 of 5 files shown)


TheoBrigitte (commented):
Isn't it simpler to replace the label in the up metric ?

QuentinBisson:
It is simpler but then we loose the WC provider information

QuantumEnigmaa (approved):
Can't think of a better solution right now, so let's try this one out

QuentinBisson:
@TheoBrigitte actually you're right, my concern is irrelevant because the first metric return an incorrect provider anyway

QuentinBisson:
I will let you review this again @giantswarm/team-atlas

TheoBrigitte (approved):

PR #1340, opened 2024-09-02
add MimirHPANeedsToBeScaledUp alert

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/31385

This PR adds `MimirHPANeedsToBeScaledUp` alert in order to detect when an HorizontalPodAutoscaler in the mimir namespace has reached its maximum capacity; meaning replicas are maxed out and resources usage is above targets.

For this alert I used the `kube_horizontalpodautoscaler_*` metrics provided by Kubernetes:
* first part of the query detects when the current desired replicas has reached the maximum allowed replicas
* second part of the query detects when the current metrics value used for autscaling (cpu or memory) is above target. When both conditions are met the alert fires.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `MimirHPANeedsToBeScaledUp` alert, to detect when Mimir's HPAs have reached maximum capacity.
+- Add `MimirHPAReachedMaxReplicas` alert, to detect when Mimir's HPAs have reached maximum capacity.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `MimirHPANeedsToBeScaledUp` alert, to detect when Mimir's HPAs have reached maximum capacity.
+
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `MimirHPANeedsToBeScaledUp` alert, to detect when Mimir's HPAs have reached maximum capacity.
```
QuentinBisson, line 12: Can we align names with https://github.com/giantswarm/prometheus-rules/blob/5276895723d59b94b7a083b80a2ced0d93be76e6/helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml#L111 ?

QuentinBisson: Maybe the alert should be the same?

TheoBrigitte:
@QuantumEnigmaa should we change this to make it work for Loki also ?

QuantumEnigmaa:
I think we can do that later. Let's merge this one and the loki one first so that we have those enabled asap and then let's create an issue to refactor this so that we can discuss how we want to do it

CHANGELOG.md
TheoBrigitte: Renamed it to `MimirHPAReachedMaxReplicas`

QuantumEnigmaa (commented):
Shouldn't you add UTs ?

QuentinBisson (approved):
LGTM apart from missing tests :)

PR #1375, opened 2024-09-24
add new loki compactor failing alert

QuentinBisson:
Towards: https://github.com/giantswarm/roadmap/issues/3683

This pull request introduces a new alert for Loki compaction. The alert is designed to notify when the compaction process in Loki did not successfully ran in the last 2 hours. This is crucial for maintaining the performance and reliability of our logging infrastructure.

### Checklist

- [x] Update CHANGELOG.md


test/tests/providers/global/platform/atlas/alerting-rules/loki.rules.test.yml
```diff
@@ -227,3 +227,35 @@ tests:
               opsrecipe: loki/
       - alertname: LokiHpaReachedMaxReplicas
         eval_time: 515m 
+
+  # Test for LokiCompactorFailedCompaction alert
+  - interval: 1m
+    input_series:
+      - series: 'loki_compactor_apply_retention_last_successful_run_timestamp_seconds{cluster_id="golem", installation="golem", namespace="loki", pipeline="testing", provider="capa"}'
+        values: "0x240 14400+60x100" # time() returns eval_time so we return 0 for the first 4 hours and then 14400 (240 minutes) for the next 100 minutes
```
hervenicol, line 235: The comment is very misleading to me.
```suggestion
        values: "0x240 14400+60x100" # compactions don't work for the first 240 minutes so the timestamp stays still, then it's updated after 240 minutes to a valid timestamp (which is number of seconds since start for the test).
```

test/tests/providers/capi/capa-mimir/platform/atlas/alerting-rules/mimir.rules.test.yml
```diff
@@ -389,7 +389,7 @@ tests:
     input_series:
       # mimir-ingester real memory usage gradually decreases until it goes below 30% of the memory requests.
       - series: 'cortex_compactor_runs_failed_total{reason="error", installation="golem", cluster_id="golem", namespace="mimir", pipeline="testing", provider="capa"}'
```
hervenicol, line 391: Not related to this PR but the comment looks completely unrelated to the metrics!

test/tests/providers/capi/capa-mimir/platform/atlas/alerting-rules/mimir.rules.test.yml
```diff
@@ -389,7 +389,7 @@ tests:
     input_series:
       # mimir-ingester real memory usage gradually decreases until it goes below 30% of the memory requests.
```
QuentinBisson, line 390:
```suggestion
```

test/tests/providers/capi/capa-mimir/platform/atlas/alerting-rules/mimir.rules.test.yml
QuentinBisson: Removed it

hervenicol (dismissed):
Fine with me.

QuentinBisson:
I'll merge and release tomorrow because this will page :D

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
```diff
@@ -135,3 +135,22 @@ spec:
...
         topic: observability
+  - name: loki.compactor
+    rules:
+    - alert: LokiCompactorFailedCompaction
+      annotations:
+        dashboard: loki-retention/loki-retention
+        description: 'Loki compactor has been failing compactions for more than 2 hours.'
+        opsrecipe: loki#lokicompactorfailedcompaction
+      # This alert checks if Loki's the last successful compaction run is older than 2 hours
+      expr: (time() - loki_compactor_apply_retention_last_successful_run_timestamp_seconds > 60 * 60 * 2)
```
QuentinBisson, line 146:
```suggestion
      expr: (time() - loki_boltdb_shipper_compact_tables_operation_last_successful_run_timestamp_seconds > 60 * 60 * 2)
```

According to upstream we should use compaction metric instead of the rétention.

test/tests/providers/global/platform/atlas/alerting-rules/loki.rules.test.yml
```diff
@@ -227,3 +227,35 @@ tests:
               opsrecipe: loki/
       - alertname: LokiHpaReachedMaxReplicas
         eval_time: 515m 
+
+  # Test for LokiCompactorFailedCompaction alert
+  - interval: 1m
+    input_series:
+      - series: 'loki_compactor_apply_retention_last_successful_run_timestamp_seconds{cluster_id="golem", installation="golem", namespace="loki", pipeline="testing", provider="capa"}'
```
QuentinBisson, line 234:
```suggestion
      - series: 'loki_boltdb_shipper_compact_tables_operation_last_successful_run_timestamp_seconds{cluster_id="golem", installation="golem", namespace="loki", pipeline="testing", provider="capa"}'
```

test/tests/providers/global/platform/atlas/alerting-rules/loki.rules.test.yml
```diff
@@ -227,3 +227,35 @@ tests:
               opsrecipe: loki/
       - alertname: LokiHpaReachedMaxReplicas
         eval_time: 515m 
+
+  # Test for LokiCompactorFailedCompaction alert
+  - interval: 1m
+    input_series:
+      - series: 'loki_boltdb_shipper_compact_tables_operation_last_successful_run_timestamp_seconds{cluster_id="golem", installation="golem", namespace="loki", pipeline="testing", provider="capa"}'
+        values: "0x240 14400+60x100" # compactions don't work for the first 240 minutes so the timestamp stays still, then it gets continuously updated after 240 minutes to a valid timestamp (which is number of seconds since start for the test).
```
QuentinBisson, line 235:
```suggestion
        values: "1x240 14400+60x100" # compactions don't work for the first 240 minutes so the timestamp stays still, then it gets continuously updated after 240 minutes to a valid timestamp (which is number of seconds since start for the test).
```

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
```diff
@@ -135,3 +135,39 @@ spec:
...
+      labels:
+        area: platform
+        cancel_if_cluster_status_creating: "true"
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "true"
+        severity: page
+        team: atlas
+        topic: observability
+    - alert: LokiCompactorFailedCompaction
```
hervenicol, line 157: Both alerts have the same name. Does it work? Won't we get into trouble at some point?

QuentinBisson: There's no reason why 2 alerts can have the same name like https://github.com/grafana/mimir/blob/2a0738f7b187664194a3af42c9b700a79d27d179/operations/mimir-mixin-compiled/alerts.yaml#L876C18-L876C63 and https://github.com/grafana/mimir/blob/2a0738f7b187664194a3af42c9b700a79d27d179/operations/mimir-mixin-compiled/alerts.yaml#L890 :)

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
```diff
@@ -135,3 +135,39 @@ spec:
...
+      annotations:
+        dashboard: loki-retention/loki-retention
+        description: 'Loki compactor has been failing compactions for more than 2 hours since start-up.'
+        opsrecipe: loki#lokicompactorfailedcompaction
+      # This alert checks if Loki's first compaction successfully ran it's first compaction compaction run is older than 2 hours
```
hervenicol, line 162: I don't understand this comment.

QuentinBisson: Well me neither :(

QuentinBisson:
```suggestion
      # This alert checks if Loki successfully ran it's first compaction less than 2 hours after it's startup
```

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
```diff
@@ -135,3 +135,39 @@ spec:
...
+        cancel_if_cluster_status_deleting: "true"
+        cancel_if_cluster_status_updating: "true"
+        cancel_if_outside_working_hours: "true"
+        severity: page
+        team: atlas
+        topic: observability
+    - alert: LokiCompactorFailedCompaction
+      annotations:
+        dashboard: loki-retention/loki-retention
+        description: 'Loki compactor has been failing compactions for more than 2 hours since start-up.'
```
hervenicol, line 160: I don't understand why we need this second alert.
What's the difference between "the compactor already ran successfully, in the past but failed for the past 2 hours" and "the compactor was never successful for the last 2 hours" ? Why do we care about the difference?

QuentinBisson: The issue is that when loki starts up (restarts are included), it's possible it does not need to run any compaction in the first few hours of it's lifetime. In that case the value of  `loki_boltdb_shipper_compact_tables_operation_last_successful_run_timestamp_seconds` will be 0.

That means that we will receive an alert after 1h (instead of 2) because the alert query would become `(time() - 0) > 2h)` => `time() > 2h` and time() will always greater than 2 hours.

hervenicol: oh, understood :bulb:

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
hervenicol: I would explain a completely different way, like:
```
This alert covers the special case at compactor startup, where the "normal" alert would always consider time `0` is more than 2 hours ago, yet we want to let it 2 hours + `for` duration.
```

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
hervenicol: Oh, I didn't know. We probably could have split some other alerts like that, rather than doing some convoluted `or` list of queries.

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
```diff
@@ -135,3 +135,39 @@ spec:
...
+        opsrecipe: loki#lokicompactorfailedcompaction
+      # This alert checks if Loki successfully ran it's first compaction less than 2 hours after startup
```
QuentinBisson, line 162:
```suggestion
      # This alert covers the special case at compactor startup, where the "normal" alert would always consider time `0` is more than 2 hours ago, yet we want to let it 2 hours + `for` duration.
```
cc @hervenicol

helm/prometheus-rules/templates/platform/atlas/alerting-rules/loki.rules.yml
QuentinBisson: I fixed it :)

hervenicol (approved):
LGTM except for some small grammar and nitpicking in a comment.

QuentinBisson:
I think they're all fixed, thanks for the review ;)

PR #1393, opened 2024-10-14
Alert StatefulsetNotSatisfiedAtlas alert

hervenicol:
Towards: https://github.com/giantswarm/roadmap/issues/3719

This PR an alert when a deployment has not all the expected pods running.

Related opsrecipe change here: https://github.com/giantswarm/giantswarm/pull/31903

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/platform/atlas/alerting-rules/statefulset.management-cluster.rules.yml
+++ b/helm/prometheus-rules/templates/platform/atlas/alerting-rules/statefulset.management-cluster.rules.yml
@@ -1,0 +1,33 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  creationTimestamp: null
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+{{- if not .Values.mimir.enabled }}
+    cluster_type: "management_cluster"
+{{- end }}
+  name: deployment.management-cluster.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+  - name: statefulset
+    rules:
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/platform/atlas/alerting-rules/statefulset.management-cluster.rules.yml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  groups:
+  - name: statefulset
+    rules:
+    - alert: StatefulsetNotSatisfiedAtlas
+      annotations:
+        description: '{{`Statefulset {{ $labels.namespace}}/{{ $labels.statefulset }} is not satisfied.`}}'
+        opsrecipe: deployment-not-satisfied/
+      expr: |-
+        kube_statefulset_status_replicas{cluster_type="management_cluster", statefulset=~"loki.*|mimir.*"}
```
QuentinBisson, line 21: Maybe we use namespace instead lf statefulset? The values would be the same but we would not need to use regex?

hervenicol: I wanted to stick to the current `deploymentnotsatisfied` alert's behaviour.

I guess the reason why it's based on the pod name is for apps running in `kube-system`, where we have to filter by pod name to then route to the proper team.

So, for today/current case, filtering by namespace would work, but if we want to add a kube-system statefulset to the alert we would have to revert.

QuentinBisson: Sure :)

PR #1396, opened 2024-10-16
Update KyvernoWebhookHasNoAvailableReplicas alert to target resource webhooks specifically

fhielpos:
I have seen this alert flipping for a while when Kyverno is downscaled (admission-controller) but the rest of the components are still running, which causes the alert to trigger.

Ideally, we don't care about the other webhooks since the one that would block actual workloads from the cluster is the resource one. Since this alert only checks for kyverno or kyverno-admission-controller deployment, it only makes sense to monitor that webhook.


stone-z (approved):
Great idea

PR #1397, opened 2024-10-21
Add InhibitionClusterWithoutWorkerNodes for CAPA

hervenicol:
Towards: https://github.com/giantswarm/giantswarm/issues/31390

This PR adds an inhibition for clusters that have no worker nodes.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
+++ b/helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
@@ -1,0 +1,34 @@
+{{- if eq .Values.managementCluster.provider.kind "capa" }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  creationTimestamp: null
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+    cluster_type: "management_cluster"
+  name: capa.inhibitions.rules
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+  - name: capa.inhibitions
+    rules:
+    - alert: InhibitionClusterWithoutWorkerNodes
+      annotations:
+        description: '{{`Cluster ({{ $labels.cluster_id }}) has no worker nodes.`}}'
+      expr: |-
+        label_replace(
+            capi_cluster_status_condition{type="ControlPlaneReady", status="True"},
+                "cluster_id",
+                "$1",
+                "name",
+                "(.*)"
+            ) == 1
+        unless on (cluster_id) (
+            sum(capi_machinepool_spec_replicas{} > 0) by (cluster_id)
+        )
+      labels:
+        area: kaas
+        has_worker_nodes: "false"
+        team: phoenix
+        topic: status
+{{- end }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added InhibitionClusterWithoutWorkerNodes for CAPA
+
 ### Changed
 
 - Modify `KyvernoWebhookHasNoAvailableReplicas` to check specifically for Kyverno resource webhook.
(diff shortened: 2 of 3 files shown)


helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: {{ .Values.namespace  }}
+spec:
+  groups:
+  - name: capa.inhibitions
+    rules:
+    - alert: InhibitionClusterWithoutWorkerNodes
+      annotations:
+        description: '{{`Cluster ({{ $labels.cluster_id }}) has no worker nodes.`}}'
+      expr: |-
+        label_replace(
```
QuentinBisson, line 19: Why is this for capa only? otherwise it looks good

helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
```diff
@@ -0,0 +1,34 @@
...
+            capi_cluster_status_condition{type="ControlPlaneReady", status="True"},
+                "cluster_id",
+                "$1",
+                "name",
+                "(.*)"
+            ) == 1
+        unless on (cluster_id) (
+            sum(capi_machinepool_spec_replicas{} > 0) by (cluster_id)
```
QuentinBisson, line 27: Do we accept 1 worker nodes only?

helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
QuentinBisson: I would think that with capi_machinepool_spec_replicas and `capi_machinedeployment_spec_replicas` we should be good?

hervenicol: Because the metrics is only for CAPA.
Ref: https://gigantic.slack.com/archives/C02HLSDH3DZ/p1729502169569249?thread_ts=1729498455.102279&cid=C02HLSDH3DZ

helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
hervenicol: Prometheus-agent/Alloy could run on a 1-node WC, so that's potentially ok.

helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
QuentinBisson: Makes sense :)

QuentinBisson (approved):

helm/prometheus-rules/templates/kaas/phoenix/alerting-rules/capa.inhibition.rules.yml
fiunchinho: Are we sure we want to use `capi_machinepool_spec_replicas`? That's the number of replicas defined in the `MachinePool` spec, but that doesn't necessarily represent the _current_ number of nodes. For the inhibition, I thought we would prefer to use the current number of replicas. There are other metrics like

- `capi_machinepool_status_replicas`: Replicas is the most recently observed number of replicas.
- `capi_machinepool_status_replicas_ready`: The number of ready replicas for this MachinePool. A machine is considered ready when the node has been created and is "Ready".
- `capi_machinepool_status_replicas_available`: The number of available replicas (ready for at least minReadySeconds) for this MachinePool.

Do you think it would make sense to use one of those instead?

hervenicol: Current inhibition works when the cluster has been purposely scaled down.

If the cluster should have nodes but none is ready/available, I think we should get a page.
I don't think current state of CAPI monitoring manages it, so I'd rather have a "prometheus-agent down" alert than no alert in this case.

It seems to me that there's quite a gap between vintage AWS and CAPA alerts, but I don't think it's Atlas responsibility to fix it. So I went the quickest way to solving my actual issue :sweat_smile:

hervenicol: That would be ideal. But my first expectation is to not get paged when a cluster has no issues, for now :rofl:

PR #1438, opened 2024-11-22
Add alert to monitor Shield pods restarts

fhielpos:


ssyno (approved):

test/tests/providers/global/platform/shield/alerting-rules/general.rules.test.yml
```diff
@@ -0,0 +1,30 @@
...
+tests:
+  - interval: 1m
+    input_series:
+      # Kyverno validating webhooks
+      - series: 'kube_pod_container_status_restarts_total{app="kube-state-metrics", cluster_id="golem", cluster_type="workload_cluster", container="main", customer="giantswarm", endpoint="http", installation="golem", instance="100.94.2.68:8080", job="kube-state-metrics", namespace="security", node="master-0001", organization="giantswarm", pipeline="stable", pod="trivy-0", provider="capa", region="eu-west-1", service="kube-prometheus-stack-kube-state-metrics", service_priority="medium", uid="a38e8606-416d-4d6c-a7ea-2745078de330"}'
+        values: "0+1x60 60+0x30"
+    alert_rule_test:
+      # Trivy pod
+      - alertname: ShieldComponentRestartingTooOften
+        eval_time: 70m
```
hervenicol, line 13: The alert has to fail for 30min before paging.
And it starts failing at 60min.
So you shouldn't expect an alert before 91min.

fhielpos: Shouldn't it start failing before the 60 min? How does the increase work? I thought it would increase by 1 every minute, so should start failing after 6 mins

hervenicol: Oh, I read too fast, and I understand your point.

Maybe the increase expects 1 hour of having data, which is why it's not triggering yet at 36min?
Did you try keeping increasing 1 per minute for the whole 90min?

fhielpos: Still nothing 😭

test/tests/providers/global/platform/shield/alerting-rules/general.rules.test.yml
```diff
@@ -0,0 +1,30 @@
+---
+rule_files:
+  - kyverno.rules.yml
```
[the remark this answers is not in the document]
fhielpos: OHHHHHHHHH

PR #1444, opened 2024-11-27
Split kong alerts

ubergesundheit:
This PR tries to make area oncall a little more bearable until things can be fixed on other places

Split the existing `KongProductionDeploymentNotSatisfied` into prod and non-prod. The one for non-prod does not alert during nights any more


QuentinBisson (approved):
Can you rename thé non prod alert to Say nonprod 😅

PR #1462, opened 2024-12-20
Fix duplicate series in PromtailDown alert

TheoBrigitte:
Rules fails to evaluate because there are multiples pods matching the `up{container="promtail"} == 0` query

1. Failing evaluation
image

2. Multiple pods for `up` metrics
image

3. `kube_pod_info` is only available for `kube-system` namespace
image

This PR fixes the PromtailDown alert by limiting its scope to the kube-system namespace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix duplicate series in `PromtailDown` alert.
+
 ## [4.30.0] - 2024-12-10
 
 ### Added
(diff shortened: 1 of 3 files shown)


hervenicol:
Beware that some customers have some config to run monitoring tools in a dedicated namespace.
Currently only for prometheus-agents AFAIK, but may be for promtail as well someday.

TheoBrigitte:
Yes but we can't monitor those with this alert as metrics are limited to the kube-system namespace.

hervenicol:
> Yes but we can't monitor those with this alert as metrics are limited to the kube-system namespace.

Oh, on WCs yes! :+1: 
Then, good to me.

hervenicol (approved):

PR #1472, opened 2025-01-20
Add Mimir Alertmanager alerts

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/3752

Adding some Mimir Alertmanager alerts to detect failures.

This a bit of a random pick, but we need something to start with and evaluate if things are running smoothly.

Those alerts are taken from [upstream](https://github.com/grafana/mimir/blob/main/operations/mimir-mixin-compiled/alerts.yaml#L617), therefore using upstream ops recipes until we have better ones.


TheoBrigitte:
> This PR is missing the changelog entry as well as tests for the alerts.

Added a changelog entry, will add unit tests
 
> On top of that, could you add a comment explaining that those are coming from upstream like so?

Added a comment with a link to upstream
 
> Does the dashboard that is linked in the alert exists in our grafanas?

The dashboard uid is valid and points to our Mimir Alertmanager dashboard

> Stopping here but for the mimir alerts we put them all into the mimir alerts using another rule group:

I though having those alerts in another files would separate things and avoid a too long file for mimir.rules, but I moved them in there.

test/hack/bin/check-opsrecipes.sh
```diff
@@ -137,7 +137,8 @@ main() {
             fi
 
             # Let's check if the opsrecipe is in our list of existing opsrecipes
-            if ! isInArray "$opsrecipe" "${opsRecipes[@]}"; then
+            # or is a valid URL starting with http
+            if ! isInArray "$opsrecipe" "${opsRecipes[@]}" && [[ "$opsrecipe" != http* ]]; then
```
TheoBrigitte, line 141: What do we think about this ?

QuentinBisson:
I'm fine with splitting the alerts into multiple files but I would rather we have them in a folder then and that we do it for all alerts :)

QuentinBisson (approved):

PR #1483, opened 2025-01-30
fix mimir object storage low rate false positives

QuentinBisson:

---
Closes: https://github.com/giantswarm/giantswarm/issues/32419

This PR should address the false positive as it is based on an error rate when accessing the object storage. The metrics used have been taken from the `Mimir / Object Store dashboards`

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/platform/atlas/alerting-rules/mimir.rules.yml
+++ b/helm/prometheus-rules/templates/platform/atlas/alerting-rules/mimir.rules.yml
@@ -179,11 +179,11 @@ spec:
         opsrecipe: mimir/
       expr: |
         sum by (component, installation, cluster_id, cluster_type, pipeline, provider) (
-          rate(thanos_objstore_bucket_operation_failures_total{cluster_type="management_cluster", namespace=~"mimir"}[30m])
+          rate(thanos_objstore_bucket_operation_failures_total{cluster_type="management_cluster", namespace="mimir"}[30m])
         )
         /
         sum by (component, installation, cluster_id, cluster_type, pipeline, provider) (
-          rate(thanos_objstore_bucket_operations_total{cluster_type="management_cluster", namespace=~"mimir"}[30m])
+          rate(thanos_objstore_bucket_operations_total{cluster_type="management_cluster", namespace="mimir"}[30m])
         )
         > 0
       for: 1h
--- a/helm/prometheus-rules/templates/platform/atlas/alerting-rules/mimir.rules.yml
+++ b/helm/prometheus-rules/templates/platform/atlas/alerting-rules/mimir.rules.yml
@@ -172,6 +172,30 @@ spec:
         severity: page
         team: atlas
         topic: observability
+    - alert: MimirObjectStorageLowRate
+      annotations:
+        dashboard: 8280707b8f16e7b87b840fc1cc92d4c5/mimir-writes
+        description: '{{`Mimir object storage write rate is down.`}}'
+        opsrecipe: mimir/
+      expr: |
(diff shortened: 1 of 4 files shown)


helm/prometheus-rules/templates/platform/atlas/alerting-rules/mimir.rules.yml
```diff
@@ -287,35 +311,6 @@ spec:
         severity: page
         team: atlas
         topic: observability
-    - alert: MimirObjectStorageLowRate
```
QuentinBisson, line 290: This has been moved to the mimir rule group as it was under mimir.compactor :D

helm/prometheus-rules/templates/platform/atlas/alerting-rules/mimir.rules.yml
```diff
@@ -172,6 +172,30 @@ spec:
...
+        opsrecipe: mimir/
+      expr: |
+        sum by (component, installation, cluster_id, cluster_type, pipeline, provider) (
+          rate(thanos_objstore_bucket_operation_failures_total{cluster_type="management_cluster", namespace=~"mimir"}[30m])
+        )
+        /
+        sum by (component, installation, cluster_id, cluster_type, pipeline, provider) (
+          rate(thanos_objstore_bucket_operations_total{cluster_type="management_cluster", namespace=~"mimir"}[30m])
+        )
+        > 0
```
QuentinBisson, line 188: I do not think we need the unless here so I removed it :)

test/tests/providers/capi/capa/platform/atlas/alerting-rules/mimir.rules.test.yml
```diff
@@ -570,10 +570,10 @@ tests:
   # Test for MimirObjectStorageLowRate alert
   - interval: 1m
     input_series:
-      - series: 'cortex_bucket_store_sent_chunk_size_bytes_count{cluster_id="myinstall", cluster_type="management_cluster", installation="myinstall", namespace="mimir", pipeline="stable", provider="capa"}'
-        values: "_x90 1+1x90 90+0x200"
-      - series: 'capi_cluster_status_condition{cluster_id="myinstall", cluster_type="management_cluster", installation="myinstall", namespace="mimir", pipeline="stable", provider="capa", name="myinstall", type="ControlPlaneReady", status="True"}'
-        values: "1+0x380"
+      - series: 'thanos_objstore_bucket_operation_failures_total{cluster_id="myinstall", cluster_type="management_cluster", component="store-gateway", installation="myinstall", namespace="mimir", pipeline="stable", provider="capa"}'
```
hervenicol, line 573: You removed the `_x90` test, which checked the behaviour when the counters don't exist. This is usually what happens when no data has been sent since startup.
I'm thinking that may actually the specific case that was paging in the original issue.

Are we confident these counters always exist?
Does the `thanos_objstore_bucket_operations_total` counter exist if no data was successfully sent?
Does the `thanos_objstore_bucket_operation_failures_total` counter exist if there were no failures?
What does the rule do if one of these counters does not exist?

QuentinBisson: Let's pair tomorrow because my brain does not process 😂

QuentinBisson: So from this on gaggle:

image

The query itself pages whether we keep the unless or not (so this was not related to missing metrics) at all

The 2 metrics in question have had no missing metrics in the last 3 months but how confident do you need me to be that they are there on all clusters?

hervenicol (approved):

PR #1506, opened 2025-02-17
fix dashboard for alert

QuentinBisson:

---
This PR fixes the dashboard that is linked with this alert

### Checklist

- [ ] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


TheoBrigitte (approved):

QuentinBisson:
@TheoBrigitte I agree but the alert is about object store which is also targetting the compactors for instance so I think the alert should point to the Object Store dashboard and the ops-recipe should point to other useful dashboards. WDYT?

TheoBrigitte:
> @TheoBrigitte I agree but the alert is about object store which is also targetting the compactors for instance so I think the alert should point to the Object Store dashboard and the ops-recipe should point to other useful dashboards. WDYT?

Yes agreed.

PR #1519, opened 2025-02-25
add-first-loki-alert

QuentinBisson:

---
Towards: https://github.com/giantswarm/roadmap/issues/3680

This PR tries to add an inital log based alert to dectect CIDRNotAvailable events.

@fiunchinho I iwll deploy this on golem, do you think you could trigger those events there?

### Checklist

- [x] Update CHANGELOG.md


fiunchinho:
Nice! Ping me in slack when you need it and I'll trigger it

fiunchinho:
Recipe added here https://github.com/giantswarm/giantswarm/pull/32750

fiunchinho (approved):

QuentinBisson:
Log based alerting is currently blocked by https://github.com/giantswarm/shared-configs/pull/231 and https://github.com/giantswarm/observability-operator/pull/279 (using the giantswarm tenant instead of anonymous on mimir) but we also need to figure out how to add installation level labels like the pipeline

TheoBrigitte (approved):

QuentinBisson:
It's not ignored by the runbook scripts but it might bé ignored by the others that's true. I added an issue to add unit tests support but we might have to fix all ci tooling 🤔

hervenicol:
Loki alerts will generate some constant S3 traffic from loki.
With Mimir, this was a cost issue that we fixed with memcached.

When we release alerts for logs, how will we monitor that?
Do we have a kill switch in case it goes wrong?
Is it documented and communicated so anyone can leverage it, even if it goes crazy during the week-end?

QuentinBisson:
@hervenicol for the ci tooling, yes the file needs to be named rules.yml (maybe rules.yaml) works as well. If we name the file logs.rules.yaml then both pint and promtool fails so let's address those 2 ci checks later

hervenicol:
> for the ci tooling, yes the file needs to be named rules.yml (maybe rules.yaml) works as well.
I didn't understand this sentence. Could you rephrase?

> > Loki alerts will generate some constant S3 traffic from loki. With Mimir, this was a cost issue that we fixed with memcached.
> > When we release alerts for logs, how will we monitor that? Do we have a kill switch in case it goes wrong? Is it documented and communicated so anyone can leverage it, even if it goes crazy during the week-end?
> 
> Well no, because we have no clue what the load will be with just one alert even and the kill switch was not thought about when we worked on the investigation. I think it should be easy to do kill it (by disabling the ruler) but do we want to document it? I don't know. What would be your thoughts?

I think when we will notice the increase in costs we will need to act quick.
I'm not sure it requires oncall documentation though, but we need to know how we can shut it down globally.
So I'd love it to be documented somewhere yes.

I think disabling the ruler is a good idea, as:
- we can do it globally via shared values
- we can turn it on on specific installations for tests via user-values

QuentinBisson:
So our current alert ci tooling triggers only if the file names ends with the rules.yml suffix nothing more, nothing less and that is not something we documented explicitely

I tried to rename this specific rule to be logs.rules.yml and saw both pint and promtool tests failing

hervenicol (approved):
Congrats for the first loki rule! :tada:

PR #1355, opened 2024-09-10
add MimirContinuousTestFailingOnWrites and MimirContinuousTestFailing…

QuantumEnigmaa:
Towards: https://github.com/giantswarm/roadmap/issues/3578

As discussed in the above issue, this PR adds 2 alerts based on the mimir's `continous-test` component's metrics so that we get alerted when something is wrong in the read or write path. 

Those 2 alerts are directly taken from upstream mixins as explained in the comments. 

Before merging this PR, I still need to add UTs as well as create a dedicated dashboard for the continous-test component.

### Checklist

- [x] Update CHANGELOG.md
- [ ] Add [Unit tests](https://github.com/giantswarm/prometheus-rules/#testing)
- [x] Follow [Alert structure](https://github.com/giantswarm/prometheus-rules/#how-alerts-are-structured)
- [x] Consider [creating a dashboard](https://docs.giantswarm.io/getting-started/observability/grafana/custom-dashboards/) ([guidelines](https://intranet.giantswarm.io/docs/product/ux/guidelines/dashboards/)) (if it does not exist already) to help oncallers monitor the status of the issue.
- [ ] Request review from oncall area, as well as team (e.g: `oncall-kaas-cloud` GitHub group).

PR #1530, opened 2025-03-06
Update StatefulsetNotSatisfiedAtlas alert

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/32771

This PR updated the `StatefulsetNotSatisfiedAtlas` alert to page after `3 days`.


QuantumEnigmaa:
I think there is still one use case where this alert is useful though : I got paged by this alert when a `loki-write` was unable to schedule and was stuck in `Pending` state on `grouse` because of a lack of resources in the cluster. 
This was the only alert that paged, so if it didn't exist I wouldn't have known that this replica was unable to schedule until a more serious problem would arise.

TheoBrigitte:
I would argue that if there's no serious problem that arise I would not worry about a loki-write pod not being scheduled.

QuantumEnigmaa:
>  if there's no serious problem that arise I would not worry about a loki-write pod not being scheduled.

I guess this is true in most cases but I would say that it is also our responsability to provide a healthy observability-platform working at full capacity. 

@giantswarm/team-atlas WDYT ?

TheoBrigitte:
We already have a bunch of alerts to detect if the observability platform is healthy, this alert is just noise when a pod is not being scheduled which does not necessarily translate in unavailability of the platform.

hervenicol:
This alert is not so old, and was introduced with this issue: https://github.com/giantswarm/roadmap/issues/3719

We already had loki-canary at that time (https://github.com/giantswarm/prometheus-rules/pull/1386) for one week.
Mimir continuous tests were added shortly after: https://github.com/giantswarm/prometheus-rules/pull/1355

I wonder if these alerts work, because I reckon I've experienced broken queries via grafana, because of Mimir ingesters ring broken, but no MimirContinuousTest alerts.

Maybe we should play on alerts sensitivity?

QuentinBisson:
Maybe we should bé paged if it's pending for 1-3 days?

TheoBrigitte:
> I wonder if these alerts work, because I reckon I've experienced broken queries via grafana, because of Mimir ingesters ring broken, but no MimirContinuousTest alerts.
> 
> Maybe we should play on alerts sensitivity?

We do have `LokiRingUnhealthy` and might need one for Mimir or adjust `MimirContinuousTest`, but directly checking for stateful state seems wrong to me and will likely produce noise.

QuentinBisson:
I think this might be a good weekly reporting stuff that we talked about right?

QuentinBisson:
So maybe we get paged if it's pending for too long without urgency and then we make those reporting alerts only later?

QuantumEnigmaa:
> So maybe we get paged if it's pending for too long without urgency and then we make those reporting alerts only later?

Could be a good idea.

TheoBrigitte:
Here we go I updated the alert to page after `3 days`, can I have another round of review please ? @giantswarm/team-atlas

QuantumEnigmaa (approved):
LGTM

PR #1531, opened 2025-03-06
Remove PrometheusOperatorSyncFailed and PrometheusOperatorReconcileErrors

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/32767

Remove noisy alerts about prometheus operator reconciliations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Remove `PrometheusOperatorSyncFailed` alert.
+- Remove `PrometheusOperatorReconcileErrors` alert.
+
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/platform/atlas/alerting-rules/statefulset.rules.yml
```diff
@@ -1,33 +0,0 @@
-apiVersion: monitoring.coreos.com/v1
```
QuentinBisson, line 1: Should this be here?

TheoBrigitte: no, fixed

QuentinBisson (commented):
LGTM but it would be better with a changelog :D

QuentinBisson (approved):

PR #1559, opened 2025-03-26
Add tenant label to all alerts

QuentinBisson:

---
Towards: https://github.com/giantswarm/roadmap/issues/3949

This PR adds the giantswarm tenant label to all giantswarm alerts.

It also gets rid of the alloy config as it is now taken care of by the observability operator

### Checklist

- [x] Update CHANGELOG.md
- [x] Consider [creating a dashboard](https://docs.giantswarm.io/tutorials/observability/data-exploration/creating-custom-dashboards/) ([guidelines](https://intranet.giantswarm.io/docs/product/ux/guidelines/dashboards/)) (if it does not exist already) to help oncallers monitor the status of the issue.


helm/prometheus-rules/templates/alloy-rules.yaml
hervenicol, line 1: Why not moving the whole alloy-rules app to olly-op then?

I'm not comfortable with this situation:
- alloy-rules app is deployed by `prometheus-rules` with no config or a default config? what does it do when it has no config given?
- then olly-op deploys the config and updates the alloy-rules app, which will redeploy with the proper config

Whereas when it was all managed here, it happened all at once.
Hence my proposal to also move the app resource to olly-op.

Another option would be to keep the configMap config in the app, which would give:
- alloy-rules app is created with a config that does not exist - so the deployment is failing
- olly-op creates the config
- alloy-rules can deploy with a proper config

hervenicol:
The fact that some code moves from prometheus-rules to olly-op makes me wonder how the migration will happen.
Should we release one app before the other?

helm/prometheus-rules/templates/alloy-rules.yaml
QuentinBisson: Well Théo did not want me to during refinement :D

QuentinBisson: Now, I would think we could enabled this as part of the alloy-metrics that we already have running to reduce the number of alloys we have running.

QuentinBisson:
We need to deploy the rules first otherwise they are deleted from mimir

QuentinBisson:
App création was moved to the operator

README.md
```diff
@@ -103,7 +103,9 @@ spec:
 These alerts are generated from log data using LogQL queries processed by Loki. To create a log-based alert:
 
 - Ensure the `expr` field contains a valid LogQL query
-- Name the file with a `.logs.yaml` extension, this will render the following label `application.giantswarm.io/prometheus-rule-kind: loki` on the alert and ensure the alert is loaded into Loki.
+- Name the file with a `.logs.yaml` extension, this will render the following labels on the alert and ensure the alert is loaded into Loki:
+  * `application.giantswarm.io/prometheus-rule-kind: loki` (deprecated, will be removed once all management clusters have been upgraded to v30)
+  * `observability.giantswarm.io/rule-type: logs` (new label for releases > v30)
```
TheoBrigitte, line 108: Do we support both label for the time being ?

QuentinBisson: Until all our MCs are on v30, this is ignored

TheoBrigitte: Ok :+1:

TheoBrigitte (approved):

fiunchinho (approved):

PR #1588, opened 2025-05-06
Add new `ClusterUpgradeStuck` alert

njuettner:
Towards: https://github.com/giantswarm/giantswarm/issues/33398

This alert checks in case a cluster app is stuck via `status="already-exists"`.

### DIFF
--- a/helm/prometheus-rules/templates/platform/honeybadger/alerting-rules/app.rules.yml
+++ b/helm/prometheus-rules/templates/platform/honeybadger/alerting-rules/app.rules.yml
@@ -200,3 +200,27 @@ spec:
         severity: page
         team: honeybadger
         topic: releng
+    - alert: ClusterUpgradeStuck
+      annotations:
+        description: '{{`Cluster upgrade for {{ $labels.app }} is stuck. Version mismatch detected for {{ $labels.name }} in namespace {{ $labels.namespace }}.`}}'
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/platform/honeybadger/alerting-rules/app.rules.yml
```diff
@@ -200,3 +200,27 @@ spec:
         severity: page
         team: honeybadger
         topic: releng
+    - alert: ClusterUpgradeStuck
+      annotations:
+        description: '{{`Cluster upgrade for {{ $labels.app }} is stuck. Version mismatch detected for {{ $labels.name }} in namespace {{ $labels.namespace }}.`}}'
+        runbook_url: https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/app-pending-update/
+      expr: |-
+        (
+          app_operator_app_info{catalog="cluster", app=~"cluster-.*", version_mismatch="true", status="already-exists"}
```
AverageMarcus, line 209: Just to double check - this wont trigger on other apps starting with `cluster-` right? Because the `cluster` catalog only has the provider charts in?

QuentinBisson (approved):

helm/prometheus-rules/templates/platform/honeybadger/alerting-rules/app.rules.yml
njuettner: Yes 👍🏻

AverageMarcus (approved):

PR #1590, opened 2025-05-06
DeploymentNotSatisfiedAtlas: lower sensitivity and page only during business hours

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/33400

Here is a reason why we don't need to page out of business hours, for each monitored component managed by this alert:
- **alertmanager**: if it breaks we eventually have a heartbeat alert
- **grafana**: if it breaks we eventually have a `grafanaDown` alert
- **logging-operator**: not critical for night and week-end operations
- **loki**: If one component is down and the service still works we don't care. If the service breaks we have `LokiMissingLogs`.
- **mimir**: If one component is down and the service still works we don't care. If the service breaks we have heartbeats and `MimirContinuous` alerts
- **oauth2-proxy**: we don't use it anymore, let's remove it from the alert.
- **object-storage**: not critical for night and week-end operations.
- **observability-gateway**: covered by `LogReceivingErrors`
- **observability-operator**: not critical for night and week-end operations.
- **prometheus**: we don't run any `prometheus`-named deployments anymore on CAPI.
- **promxy**: we don't run promxy anymore on CAPI.
- **tempo**: we don't run tempo yet, and when we will we should cover it with higher level alerts like mimir continuous tests and loki canary.
- **pyroscope**: we don't run pyroscope yet, and when we will we should cover it with higher level alerts like mimir continuous tests and loki canary.
- **silence-operator**: not critical for night and week-end operations.
- **sloth**: not critical for night and week-end operations.

Are you all fine with that?

### Checklist

- [ x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- DeploymentNotSatisfiedAtlas: lower sensitivity and page only during business hours
+
 ## [4.57.0] - 2025-04-30
 
 ### Changed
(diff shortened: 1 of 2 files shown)


QuentinBisson:
Are we sure about the loki alert?

If we are then I'm definely fine :)

hervenicol:
I'm not totally sure about the loki alert.
But even then, does a loki outage deserve being awaken out of business hours?

QuentinBisson:
I would argue that having some kind of hearbeat alert for loki is useful but probably not

QuentinBisson (approved):

hervenicol:
> I would argue that having some kind of hearbeat alert for loki is useful but probably not

I Agree, and that's what the `LokiMissingLogs` alert does, via the canary pod. My concern about this one is that I can't remember if it ever paged. Probably because we have `DeploymentNotSatisfied` way before.

Also, this PR does not remove the DeploymentNotSatisfied alert: it only delays it, and prevents it from paging at night. So we are still going to be informed when these components fail, but during business hours and with a longer delay.

QuantumEnigmaa (approved):

PR #1594, opened 2025-05-07
Add IncorrectResourceUsageData alert

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/33350

This PR adds the IncorrectResourceUsageData alert to detect when data used in the Resource Usage dashboard in Grafana Cloud is incorrect.

It does so by comparing `aggregation:node:cpu_cores_total` and `machine_cpu_cores` metrics, and fire when the former has more than 10% difference with the latter. Same for memory.

The use of `quantile_over_time` and 120:30m window help smoothen the data to avoid false positives.

**Question: Should this be an alert own by Atlas and Phoenix ?**

Example:

Tested on `violet` and compared this against the values in the Resource Usage dashboard, this matches with the "gaps".

image

<details>
<summary>other installations</summary>

Looks correct, should not page much. `cedar` is having issues but usage package is below 1 so all good.

alba - data shown there match with time when we had issues with mimir
image

agama
image

alligator
image

anemone
image

armadillo
image

avocet
image

cedar
image

enigma
image

finch
image

leopard
image

sardine
image

tamarin
image

viking
image

wallaby
image

</details>


helm/prometheus-rules/templates/platform/atlas/alerting-rules/grafana-cloud.rules.yml
```diff
@@ -29,6 +29,23 @@ spec:
         severity: page
         team: atlas
         topic: observability
+  - name: resource-usage
+    rules:
+    - alert: IncorrectResourceUsageData
+      annotations:
+        description: '{{`Data used in the Grafana Cloud Resource Usage dashboard is incorrect for cluster {{ $labels.cluster_id }}.`}}'
```
QuentinBisson, line 36: Could you add some unit tests?

hervenicol, line 36: Actually the opsrecipe explains a lot!

```suggestion
                # More info about this alert can be found in the opsrecipe: https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/resource-usage-dashboard/
                description: '{{`Data used in the Grafana Cloud Resource Usage dashboard is incorrect for cluster {{ $labels.cluster_id }}.`}}'
```

TheoBrigitte: Isn't this obvious as there's already the ops-recipe link below ?

hervenicol: Didn't seem obvious to me that the opsrecipe would not only explain how to fix the issue, but also how the alert is written, and why we consider it's an important issue.

Maybe that should be an opsrecipe standard, but AFAIK very few of ours do that :sweat_smile:

QuentinBisson: I think adding a comment on top of the alert that explain what is it and how it works makes sense for all alerts though, because going to the ops-recipe all the time is not the most straigthforward access to the information

TheoBrigitte: Added the unit tests

TheoBrigitte: Added the comment

QuentinBisson (approved):

njuettner (approved):

PR #1613, opened 2025-05-12
Fix grafana cloud unit tests

TheoBrigitte:
Related: https://github.com/giantswarm/prometheus-rules/pull/1594

Just found out that grafana-cloud unit test where not running due to incorrect files naming. So I fixed those and the failing tests afterwards.


QuentinBisson (approved):

QuentinBisson:
Thinking about it, I think we should make this a global test instead of a capz and capa test

hervenicol:
> Thinking about it, I think we should make this a global test instead of a capz and capa test

I plan to do some cleanup there someday.

QuentinBisson:
I'll clean up in another pr

PR #1632, opened 2025-05-20
UTs: remove capi folder

hervenicol:
While investigating a bug with cloud-director tests, I found that it was caused by a regex in `template-chart.sh`.
Turns out this regex is not needed anymore now that we stopped supporting vintage.
So I removed the capi folder, simplified the tree, and simplified the script.

### Checklist

- [x] Update CHANGELOG.md


hervenicol:
As pint was broken, I noticed it was not dynamically reading list of cloud-providers.
So I updated the pint script to do it so it gets fixed.

QuentinBisson (approved):

PR #1646, opened 2025-06-06
Improve ClusterAutoscalerFailedScaling alert to reduce false positives

njuettner:
Towards: https://github.com/giantswarm/giantswarm/issues/33440

This PR improves the `ClusterAutoscalerFailedScaling` alert to reduce false positives by combining two expressions:
- More than 3 failures in the last 15 minutes AND
- Recent failures in the last 5 minutes (ensuring it's an ongoing issue)

### DIFF
--- a/helm/prometheus-rules/templates/kaas/tenet/alerting-rules/cluster-autoscaler.rules.yml
+++ b/helm/prometheus-rules/templates/kaas/tenet/alerting-rules/cluster-autoscaler.rules.yml
@@ -28,8 +28,7 @@ spec:
       annotations:
         description: '{{`Cluster-Autoscaler on {{ $labels.cluster_id }} has failed scaling up {{ $value | printf "%.0f" }} times recently.`}}'
         runbook_url: https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/cluster-autoscaler-scaling/
-      expr: increase(cluster_autoscaler_failed_scale_ups_total{provider=~"capa|capz|eks"}[15m]) > 3 and rate(cluster_autoscaler_failed_scale_ups_total{provider=~"capa|capz|eks"}[5m]) > 0
-      for: 5m
+      expr: increase(cluster_autoscaler_failed_scale_ups_total[15m]) > 1 and rate(cluster_autoscaler_failed_scale_ups_total[5m]) > 0
       labels:
         area: kaas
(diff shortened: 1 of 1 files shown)


helm/prometheus-rules/templates/kaas/tenet/alerting-rules/cluster-autoscaler.rules.yml
```diff
@@ -26,13 +26,14 @@ spec:
         topic: cluster-autoscaler
     - alert: ClusterAutoscalerFailedScaling
       annotations:
-        description: '{{`Cluster-Autoscaler on {{ $labels.cluster_id }} has failed scaling up.`}}'
+        description: '{{`Cluster-Autoscaler on {{ $labels.cluster_id }} has failed scaling up {{ $value | printf "%.0f" }} times recently.`}}'
         runbook_url: https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/cluster-autoscaler-scaling/
-      expr: cluster_autoscaler_failed_scale_ups_total{provider=~"capa|capz|eks"} > 0
-      for: 15m
+      expr: increase(cluster_autoscaler_failed_scale_ups_total{provider=~"capa|capz|eks"}[15m]) > 3 and rate(cluster_autoscaler_failed_scale_ups_total{provider=~"capa|capz|eks"}[5m]) > 0
```
AverageMarcus, line 31: These values aren't right I don't think. We had a legit issue that this alert should have caught with `cicdprod` on the 27th but this doesn't catch it.

njuettner: Ok thanks I'll investigate

Gacko (approved):

helm/prometheus-rules/templates/kaas/tenet/alerting-rules/cluster-autoscaler.rules.yml
njuettner: image
It might be changing it to  > 1 would fire it properly WDYT?

njuettner: Yeah removed `for`, dropped the provider and adjusted the increase

AverageMarcus (approved):

PR #1676, opened 2025-07-08
Make KubeStateMetricsDown page only during business hours

hervenicol:
As discussed during Atlas alerts review, this alert should not page during the night.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


hervenicol:
> I think the main reason why it is paging outside of business hours is due to our alerts relying on it's metrics. Could you check if the alerts we have that rely on ksm are in fact not paging outside business hours?

If some alerts rely on these, they are very probably inhibited. At least I remember we did an extensive inhibition work around alerts caused by KSMDown.
Do you think we may have some serious issues during the night, that we don't see because a KSMDown alert would have paged us?

QuentinBisson (approved):

PR #1682, opened 2025-07-17
Add `ClusterControlPlaneMachineStatusNotHealthy` alert (very broad, using `severity: notify` for testing how often it happens)

AndiDog:
Towards https://github.com/giantswarm/giantswarm/issues/33868

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


iuriaranda (commented):
Won't this be super noisy during cluster rollouts, having it on a 1m period?

AndiDog:
> Won't this be super noisy during cluster rollouts, having it on a 1m period?

I think it might be, but the problem is that the problem metric is only `> 0` for a very short time – judging from the recent upgrade issue and its metrics. Not sure how to do it better right now. Maybe sum up over 15 minutes or so?

PR #1702, opened 2025-08-06
add ObservabilityAgentsDown alerting rule

QuantumEnigmaa:
Towards: https://github.com/giantswarm/giantswarm/issues/33709

This PR adds the `ObservabilityAgentsDown` alerting rule that will page only when both monitoring and logs are down on the same WC.

It will also inhibs the `MonitoringAgentDown` and `LoggingAgentDown` alerts to avoid too many alerts from paging.

### Checklist

- [x] Update CHANGELOG.md


hervenicol:
What is the purpose of this alert?
Its description says it detects if both metrics and log monitoring agents fail at the same time. But we already have some alerts for these individual issues.
The real purpose of this alert is to determine if a cluster is completely broken, right? And we're doing it this way because these are all our monitoring touchpoints to the clusters, right?
So that's what should be in the comments / description / summary of the alerts IMO.

And second point:
We determined this alert detects if a WC is completely fucked up.
Is it really Atlas we want to alert in this case?

hervenicol:
You can use the `{{ providerTeam }}` variable like on this alert: https://github.com/giantswarm/prometheus-rules/blob/main/helm/prometheus-rules/templates/kaas/tenet/alerting-rules/pods.rules.yml

QuantumEnigmaa:
Hey @giantswarm/team-tenet , as this alert will page KaaS, would you maybe prefer having it defined in your folder ?

Gacko (changes requested):
Why is this going to page KaaS? Where has this been discussed? If Atlas' components are not working as expected, then Atlas should investigate them.

Gacko:
I get your point, but then your using "all Atlas components are broken" as an indicator for "WC seems to be broken" - is there no better way for telling that? Because people will get confronted with "all Atlas components are broken" and be confused when they see it's the whole WC. So if possible, I'd rather improve the "is the WC working" alerting instead.

Gacko:
Please make it page for Atlas or have it check for something that's actually area KaaS responsibility. There's just one thing that's more annoying than getting paged at 3 AM: Getting paged for something you are not responsible for, so you cannot fix with confidence, or getting paged for flakes. The former you're about to implement here based on the *assumption* that this alert is always being caused by a broken WC. The latter is making my on-call experience worse lately, especially because of alerts I, my team or my area did not implement.

Gacko:
Also, from reading the expression again: This would also page if the MC Ingress is not working. Or if your components are down on purpose. And the more I think about it, the more reasons come to my mind. 🥲

PR #1717, opened 2025-08-25
grafana-cloud rules: add job scraping failures stats

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/26963

If we want a global dashboard showing failing scraping jobs, we should send data to grafana cloud.
So let's create a recording rule with this info.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/helm/prometheus-rules/templates/platform/atlas/recording-rules/grafana-cloud.rules.yml
+++ b/helm/prometheus-rules/templates/platform/atlas/recording-rules/grafana-cloud.rules.yml
@@ -574,3 +574,16 @@ spec:
             "(.*)"
           )
         record: aggregation:slo:time_period:days
+      - expr: |-
+          (
+            count(up{}== 0 )
+            by (job, installation, cluster_id, provider,
+              pipeline, cluster_type, metrics_path,
+              container, namespace, organization, team)
+            /
+            count(up{})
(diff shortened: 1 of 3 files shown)


helm/prometheus-rules/templates/platform/atlas/recording-rules/grafana-cloud.rules.yml
```diff
@@ -574,3 +574,16 @@ spec:
             "(.*)"
           )
         record: aggregation:slo:time_period:days
+      - expr: |-
+          (
+            count(up{}== 0 )
```
QuentinBisson, line 579: Why do we need the division? is the first one not enough?

QuentinBisson: Do you mind adding a comment here though?

Also, maybe we want to use this recording rule in the alert instead?

hervenicol: Comments added :heavy_check_mark: 
We can probably use the recording rule in the alert as well, but I'd like to play with the generated metrics in a larger scale before.

QuentinBisson (approved):

PR #1743, opened 2025-09-30
Add memory-pressure taint monitoring for cp nodes

njuettner:
Towards: https://github.com/giantswarm/giantswarm/issues/33896

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New alert `ControlPlaneNodeMemoryPressureTaint`.
 - New alert `IRSAClaimNotReady` to monitor Crossplane IRSA objects.
 - Add quicker alerts for Kyverno's `svc-fail` validation/mutation webhooks taking very long or timing out
+- Add alerts for EFS pods
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


helm/prometheus-rules/templates/kaas/tenet/alerting-rules/node.management-cluster.rules.yml
```diff
@@ -114,3 +114,17 @@ spec:
         severity: notify
         team: {{ include "providerTeam" . }}
         topic: infrastructure
+    - alert: ManagementClusterControlPlaneNodeMemoryPressureTaint
+      annotations:
+        description: '{{`Control plane node {{ $labels.node }} has memory-pressure taint applied, indicating memory issues.`}}'
```
QuentinBisson, line 119: So they need to be 2 diferent alerts if they are the same?

QuentinBisson (approved):
LGTM :)

helm/prometheus-rules/templates/kaas/tenet/alerting-rules/node.management-cluster.rules.yml
QuentinBisson: No worries, I think the node metrics should eventually be merged in one file as having 2 alerts is not that useful :)

PR #1785, opened 2025-10-22
update test tools

hervenicol:
This PR update test tools.
When upgrading to prometheus 3, some unit tests (`MimirRulerTooManyFailedQueries`) were failing because the aggregation window was too small, so I fixed it.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - update test tools to latest versions
+- fix a few alerts to bass latest checks
 - better tests output
 - Improve `KubeStateMetricsDown` and `KubeStateMetricsNotRetrievingMetrics` alert efficiency.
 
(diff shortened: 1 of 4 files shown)


hervenicol:
Wow, it seems latest version of pint catches some new missing labels in aggregations !
That's great, but I'll have to fix them before I merge.

QuentinBisson:
@hervenicol it might be worth taking a look at the warnings. We can probably reduce the resource usage of the ruler by improving some regex we do not need like those here https://gigantic.slack.com/archives/C02GDJJ68Q1/p1761230848581819

hervenicol:
> @hervenicol it might be worth taking a look at the warnings. We can probably reduce the resource usage of the ruler by improving some regex we do not need like those here https://gigantic.slack.com/archives/C02GDJJ68Q1/p1761230848581819

That's true, but not in the current PR.
I created an issue to track that: https://github.com/giantswarm/giantswarm/issues/34674

PR #2044, opened 2026-06-09
fix: Fix GrafanaPostgresqlRecoveryTestFailed alert

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/35274

Update the GrafanaPostgresqlRecoveryTestFailed with a new expression which detect when recovery test jobs are failing since more than 13h: 3 failing jobs in a row.

### DIFF
--- a/test/tests/providers/global/platform/atlas/alerting-rules/grafana.rules.test.yml
+++ b/test/tests/providers/global/platform/atlas/alerting-rules/grafana.rules.test.yml
@@ -104,7 +104,7 @@ tests:
     # Jobs are run every 6 hour (360 minutes), with a history limit set to 1 for both successful and failed jobs, no ttl is set so finished jobs stays until a next job with the same status replaces it, leading to failed job staying indefinitely regardless of newer successful jobs.
     # kube_job_status_start_time equals the time (in seconds) at which the job started, so each series only appears from that point on.
     input_series:
-      # A job succeeds at minute 1 and stays until the next successful job in 1440 minutes (24 hours).
+      # A job succeeds at minute 1 and stays until the next successful job at hour 24 (1440 minutes)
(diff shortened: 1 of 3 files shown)


test/tests/providers/global/platform/atlas/alerting-rules/grafana.rules.test.yml
```diff
@@ -93,23 +93,29 @@ tests:
...
-  # Test: all recovery jobs failed for over 1 day triggers alert, new successful job clears it.
+  # Test: the most recent recovery test job that failed started after the most recent one that succeeded, for over 1 day, triggers the alert. A newer successful job clears it.
   - interval: 1m
     input_series:
-      # Two failed jobs persist for 1500 minutes (~25h).
-      # At minute 1450, a new successful job appears.
-      - series: 'kube_job_failed{job_name="grafana-postgresql-recovery-test-001", namespace="monitoring", condition="true", cluster_id="golem", installation="golem", pipeline="stable", provider="aws"}'
+      # An older job succeeded (start_time 100), then a newer job failed (start_time 200) and keeps failing.
+      - series: 'kube_job_status_start_time{job_name="grafana-postgresql-recovery-test-001", namespace="monitoring", cluster_id="golem", installation="golem", pipeline="stable", provider="aws"}'
+        values: "100+0x1500"
```
hervenicol: During tests, time starts at 0.
For accuracy, a `kube_job_status_start_time` with value 100 should appear at time 100.
For example:
```suggestion
        values: "_x100 100+0x1500"
```

The current syntax sets the start time in the future.

TheoBrigitte: Good point

TheoBrigitte: I've updated 2 test to represent correct value for `kube_job_status_start_time`, but also updated the timeline so it matches our real case of Job being run every 6 hour. Hopefully the comments are self-explanatory.

hervenicol (approved):

PR #2049, opened 2026-06-11
chore(release): v4.107.1

taylorbot:


hervenicol (approved):
We have a broken dashboard that we should fix before.

hervenicol (changes requested):
We have a broken dashboard that we should change before.

hervenicol (approved):
Dumb me, I was so focused on current works on dashboards I thought this PR was for dashboards 🤦 

All good here. Sorry.

PR #2067, opened 2026-06-30
feat: add etcd quorum loss alerts

njuettner:
Towards https://github.com/giantswarm/roadmap/issues/4158

Adds `ManagementClusterEtcdQuorumLost` and `WorkloadClusterEtcdQuorumLost`, firing when a majority of etcd members report no leader (`count(has_leader==0)/count(has_leader) > 0.5`, `for: 2m`). Detects true quorum loss in ~2 minutes instead of relying on the per-member `EtcdHasNoLeader` (35m on workload clusters), while staying quiet during rolling upgrades and transient elections.

Uses an `app="etcd"` whitelist rather than a `container!="loki"` blacklist: other dskit components (loki, formerly promtail) can also expose `etcd_server_has_leader`, and for a ratio alert a stray series dilutes the denominator and could suppress a real page.

Shipped as `severity: notify` for an initial observation period before promoting to `page`. Includes unit tests (quorum loss fires; single-member loss, transient elections, and non-etcd series stay excluded). Verified against production data (alba, 75 etcd members).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `ManagementClusterEtcdQuorumLost` and `WorkloadClusterEtcdQuorumLost` alerts that fire when a majority of etcd members report no leader (quorum loss). Shipped as `severity: notify` for an initial observation period before promoting to `page`.
+
 ### Changed
 
 - Improved `GrafanaPostgresqlArchivingFailure` alert to avoid false positives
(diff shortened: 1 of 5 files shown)


Gacko (approved):

hervenicol:
> Shipped as `severity: notify` for an initial observation period before promoting to `page`

May I suggest an intermediate step with `cancel_if_outside_working_hours: "true"` so you test pages during business hours only first?

Also, as-is, with no `cancel_if_outside_working_hours` label, this alert will page 24/7 for `stable-testing` intallations. If this is not what you want, please add `cancel_if_outside_working_hours: {{ include "workingHoursOnly" . }}`

njuettner:
@hervenicol thanks! I added `cancel_if_outside_working_hours: {{ include "workingHoursOnly" . }}`

hervenicol (approved):

PR #2085, opened 2026-07-16
feat: Add a new notification alert for huge audit logs changes

pipo02mix:

* Documentation: https://intranet.giantswarm.io/docs/observability/slo-alerting/

---
Towards: https://gigantic.slack.com/archives/C01SBUAHPK7/p1783536093145299

This PR creates a new alert to detect large increases in audit logs, which most of the time come from a bad actor deployed in the cluster.

### Checklist

- [x] Update CHANGELOG.md
- [x] Consider [creating a dashboard](https://docs.giantswarm.io/tutorials/observability/data-exploration/creating-custom-dashboards/) ([guidelines](https://intranet.giantswarm.io/docs/product/ux/guidelines/dashboards/)) (if it does not exist already) to help oncallers monitor the status of the issue.


stone-z:
I don't think it makes sense to page Shield here, since the audit logs can be busy for any number of reasons (even the case you linked was a customer and maybe cabbage instance). To page correctly on volume changes, the alert would need to determine which component has the huge change and then page the owner of the component.

pipo02mix:
The problem is that we don't know what the component is generating the load. In the incident we had last time, a customer team was hammering the API; IMO, we should page based on that so we can correct the behavior.  Is @giantswarm/team-tenet responsible for the cluster API?  

I have checked the query in multiple installations and it is something very weird to page (notify)

stone-z:
tl;dr generally not sure if the alerting value exceeds the distraction cost, I'll leave it up to Tenet if they want to try it out, left some suggestions.

---

My main concern is just that this will be noisy, especially if it's all routed to one team. It's hard to define "normal", and customer workloads behave unpredictably (e.g. some have big nightly batch jobs).

> we don't know what the component is generating the load

It's an extra step, but maybe we could figure it out. The audit log contains, for example, the username and userAgent fields:
```
"kind": "Event",
  ...
  "requestURI": "/apis/reports.kyverno.io/v1/namespaces/goose-vcluster/ephemeralreports",
  "verb": "create",
  "user": {
    "username": "system:serviceaccount:kyverno:kyverno-reports-controller",
  },
  "userAgent": "reports-controller/v0.0.0 (linux/amd64) kubernetes/$Format",
```

We could change the query to be the number of requests _per username_ , for example, with something like `sum by(user_username) (rate({scrape_job="audit-logs"} | json [$__auto]))`. That yields output which is still very spiky, which is why I'm concerned about noise, but we could potentially smooth it with stats:
image

If we can then associate an owner with each ServiceAccount name (either in metrics or alerting), we can directly page the owner. There would be many we can't identify, like customer workloads. But that would at least alert on large changes in managed components.

A human would need to do this anyway in order to act on the alert (probably worth putting in a dashboard even if there's no alert).

As the owners of the API server, I agree Tenet would technically be the correct team for a catch-all alert for volume without a clear owner. I will leave it up to them if they want to try this for a time. Maybe this is ok on a `notify` basis.

Gacko (approved):

pipo02mix:
Ok, I have reworked the query to be per-user so the alert names the responsible identity instead of just the cluster. Now:

  - Grouping is now sum by (…, user) on both the current and 24h-ago rates, and the description leads with User {{ $labels.user }} on workload cluster ….
  - To handle the "bad actor freshly deployed" case (a new user has no 24h baseline, so a pure 3× comparison would miss it), the expr has two branches: existing user spiking >3× its own level 24h ago, or a brand-new user with no prior baseline — both gated by the same > 20 events/s floor to keep noise down.

helm/prometheus-rules/templates/platform/shield/alerting-rules/api-audit-logs.rules.yml
```diff
@@ -0,0 +1,67 @@
...
+                  rate(k8s_api_audit_requests_total{cluster_type="workload_cluster"}[30m] offset 1d)
+                )
+            )
+          for: 1h
+          labels:
+            area: platform
+            cancel_if_outside_working_hours: "true"
+            severity: notify
+            team: shield
+            topic: security
```
stone-z, line 67:
```suggestion
            team: tenet
            topic: kubernetes
```

test/tests/providers/global/platform/shield/alerting-rules/api-audit-logs.rules.test.yml
```diff
@@ -0,0 +1,81 @@
...
+              area: platform
+              cancel_if_outside_working_hours: "true"
+              cluster_id: "12345"
+              customer: giantswarm
+              installation: golem
+              pipeline: stable
+              provider: capa
+              severity: notify
+              team: shield
+              topic: security
```
stone-z, line 55:
```suggestion
              team: tenet
              topic: kubernetes
```

test/tests/providers/global/platform/shield/alerting-rules/api-audit-logs.rules.test.yml
```diff
@@ -0,0 +1,81 @@
...
```
pipo02mix:
```suggestion
              team: tenet
              topic: kubernetes
```

test/tests/providers/global/platform/shield/alerting-rules/api-audit-logs.rules.test.yml
```diff
@@ -0,0 +1,81 @@
...
+    alert_rule_test:
+      - alertname: WorkloadClusterAuditLogVolumeSpike
+        eval_time: 1530m
+        exp_alerts: []
+
+  # Relative spike but below the absolute floor: baseline ~1 events/s, current
+  # ~10 events/s. 10 > 3*1 but 10 < 20 -> noise guard suppresses it -> no alert.
+  - interval: 1m
+    input_series:
+      - series: 'k8s_api_audit_requests_total{cluster_id="12345", cluster_type="workload_cluster", customer="giantswarm", installation="golem", pipeline="stable", provider="capa", user="system:serviceaccount:kube-system:foo"}'
```
hervenicol: Would be better to set example cluster names here.
Pro tip: You can ask your agent to follow the brand new contributing rules here: https://github.com/giantswarm/prometheus-rules/blob/main/CONTRIBUTING.md

CHANGELOG.md
```diff
@@ -86,6 +86,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `WorkloadClusterAuditLogVolumeSpike` alert (team shield, `severity: notify`) firing when a single kube-apiserver user (ServiceAccount/identity) on a workload cluster sustains an audit event rate more than 3x above its own level 24h ago — or newly appears with no prior baseline — above an absolute floor for 1h, naming the workload responsible for runaway audit log growth.
```
stone-z:
```suggestion
- Add `WorkloadClusterAuditLogVolumeSpike` alert (team tenet, `severity: notify`) firing when a single kube-apiserver user (ServiceAccount/identity) on a workload cluster sustains an audit event rate more than 3x above its own level 24h ago — or newly appears with no prior baseline — above an absolute floor for 1h, naming the workload responsible for runaway audit log growth.
```

stone-z (approved):
Ok with me once CI + Herve are happy 😄

pipo02mix:
@giantswarm/team-tenet I am going to merge this PR. I have monitor a couple of busy clusters to ensure there are no false positives. Please let me know in case you find any trouble

hervenicol (approved):

PR #2089, opened 2026-07-21
feat(cabbage): add Hubble TLS certificate expiry and certgen job failure alerts

weatherhog:
Towards https://github.com/giantswarm/giantswarm/issues/37201

### What

- `CiliumHubbleTLSCertificateWillExpireSoon` (page, working hours): fires when any Hubble TLS cert secret (`hubble-server-certs`, `hubble-relay-client-certs`, `hubble-relay-server-certs`, `hubble-ui-client-certs`) is <30 days from expiry — covering both the leaf (`tls.crt`) and the embedded Cilium CA copy (`ca.crt`). No `cluster_type` filter, so workload clusters are covered.
- `CiliumHubbleCertificateRenewalJobFailed` (page, working hours): fires when a `hubble-generate-certs` certgen job fails. With `--ca-enforce-validity-throughout-leaves-duration` (wired in the upcoming cilium-app cronJob-method change), certgen fails ~1 year before the CA stops covering new leafs — this alert is the CA early-warning path, since certgen never rotates the CA (cilium/certgen#500).

### Why

Hubble TLS certs broke hubble-relay on alba/odp01 (leaf expiry) and alligator/deu02 (CA expiry) with zero paging: `cert_exporter_secret_not_after` series for the hubble secrets exist on WCs, but the catch-all `CertificateSecretWillExpireInLessThanTwoWeeks` is scoped to `cluster_type="management_cluster"`. `cilium-ca` itself is an Opaque secret cert-exporter can't see, hence the job-failure alert as the CA signal.

The 30d threshold cannot false-positive in steady state: the cronJob method re-issues 12-month leafs every 4 months (≥8 months headroom), and it also catches clusters still on the never-renewing `helm` method during rollout.

Runbook: giantswarm/giantswarm#37228

### Testing

- promtool unit tests for both alerts (firing + name-filter non-matching cases)
- `promtool check rules` + `pint lint` on the rendered output pass (only pre-existing-style regex warnings)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `CiliumHubbleTLSCertificateWillExpireSoon` alert (`severity: page`, working hours only) firing when a Hubble TLS certificate secret (`hubble-server-certs`, `hubble-relay-client-certs`, `hubble-relay-server-certs`, `hubble-ui-client-certs` — leaf and embedded CA) expires in less than 30 days, on both management and workload clusters. The previous catch-all secret alert only covered management clusters, so Hubble cert expiry on workload clusters broke hubble-relay unnoticed ([giantswarm#37201](https://github.com/giantswarm/giantswarm/issues/37201)).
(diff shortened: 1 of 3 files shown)


QuantumEnigmaa:
We don't have a `cilium-hubble-tls-certificates` runbook yet, hence the related failing ci

weatherhog:
> We don't have a `cilium-hubble-tls-certificates` runbook yet, hence the related failing ci

https://github.com/giantswarm/giantswarm/pull/37228 The PR is open. Its only the order that the Runbook needs to be merged first.

mcharriere (approved):

QuantumEnigmaa (approved):

PR #2129, opened 2026-08-25
feat(honeybadger): alert when a cluster's services CIDR is unreadable

weatherhog:
Towards: https://github.com/giantswarm/giantswarm/issues/37031

Adds `ClusterServiceCIDRMissing` (team honeybadger, `severity: page`) on the new `cluster_apps_operator_cluster_service_cidr_missing` metric from giantswarm/cluster-apps-operator#796.

`cluster-apps-operator` derives `clusterDNSIP` from `Cluster.spec.clusterNetwork.services.cidrBlocks[0]` and falls back to the installation-wide default when it cannot read that field. `chart-operator` runs `hostNetwork: true` with `dnsPolicy: None`, so `clusterDNSIP` is its **only** resolver — a wrong value means it cannot resolve the OCI registry and every chart pull on that cluster fails. The failure is dormant: chart-operator only needs DNS when pulling a *new* chart version, so a cluster can carry a broken value for weeks and then fail an unrelated app upgrade. That is #37031.

Following the existing drift idiom in this repo (`app_operator_app_info{version_mismatch="true"}`): the operator computes the boolean, PromQL just filters on it.

### Routing

`severity: page` (PagerDuty) with `cancel_if_outside_working_hours: "true"`, so it pages during working hours only. When it fires nothing is broken yet — chart-operator only needs DNS on the next chart pull — so there is no reason to wake anyone, but it does need a human. This matches every other honeybadger `severity: page` alert; `all_pipelines` left unset, as they all do.

### Why `for: 30m`

Rides out the cluster-creation window where the `Cluster` CR exists before `clusterNetwork` is populated. There is a unit test case for exactly that (a 10-minute blip must not alert).

### Expected noise: none

I swept every installation I have Teleport access to — **148 `Cluster` CRs across 25 installations all have `services.cidrBlocks` set**. `giantswarm/cluster` makes it required with `minItems: 1` and a default, so a chart-templated cluster always has it. This alert is silent today and fires only on a genuine regression, or on a `Cluster` CR not produced by that chart. No baseline tuning needed.

### ⚠️ Merge order

**This must merge after giantswarm/giantswarm#37493**, which adds the runbook. `check-runbooks.sh` clones `giantswarm/giantswarm@main` and validates every `runbook_url` resolves to a real file, so `test-runbooks` will fail here until that PR lands. Everything else passes.

It also needs the metric to exist, so realistically: cluster-apps-operator#796 → release → this.

### Verification

Ran locally against the templated chart (`promtool check rules`, `promtool test rules`, `pint lint`) — all pass. `make test-rules` itself needs bash 4 for `mapfile`, which macOS doesn't ship, so I replicated the harness steps by hand: `helm template` per the harness flags, `yq '.spec'` extraction, then promtool.

Confirmed the unit test isn't vacuous by inverting the rule: `for: 30m` → `5m` breaks the "no alert during the creation blip" case, and `> 0` → `>= 0` breaks the "healthy cluster never alerts" case. Both fail as they should.

### Checklist

- [x] Update CHANGELOG.md
- [x] Add unit tests
- [x] Follow alert structure
- [ ] Consider creating a dashboard — not adding one. This is a boolean that should always be `0`; the runbook's investigation path is kubectl-based and a single-stat panel would not add anything an alert plus runbook doesn't already give the oncaller.
- [ ] Request review from oncall area as well as team


CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `ClusterServiceCIDRMissing` (team honeybadger), firing when
```
hervenicol: I like my changelog messages shorter.

Usually, I ask my agent to follow these rules:
- changelog says what was changed, in 1 short line
- comments can give some context about some technical choices for some changes
- PR body explains why we had to do these changes
- changelog and comments should not contain any real cluster name / information
- changelog and comments should not contain any link to private repositories

helm/prometheus-rules/templates/platform/honeybadger/alerting-rules/cluster-apps-operator.rules.yml
```diff
@@ -0,0 +1,36 @@
...
+    - alert: ClusterServiceCIDRMissing
+      annotations:
+        description: '{{`Cluster {{ $labels.cluster_namespace }}/{{ $labels.cluster_id }} has no services CIDR, so cluster-apps-operator is emitting the installation default clusterDNSIP.`}}'
+        runbook_url: '{{`https://intranet.giantswarm.io/docs/support-and-ops/runbooks/cluster-service-cidr-missing/?INSTALLATION={{ $labels.installation }}&CLUSTER={{ $labels.cluster_id }}`}}'
+      expr: cluster_apps_operator_cluster_service_cidr_missing > 0
+      for: 30m
+      labels:
+        area: platform
+        cancel_if_cluster_control_plane_unhealthy: "true"
+        cancel_if_outside_working_hours: {{ include "workingHoursOnly" . }}
```
hervenicol: This will make your alert page out of business hours for production clusters.
Is this really what you want? (asking because I know this to be a common source of confusion)

weatherhog: Good catch, and no — that wasn't what I wanted. Changed to a hardcoded `cancel_if_outside_working_hours: "true"`.

I'd reached for the `workingHoursOnly` helper by pattern-matching on nearby alerts without checking what it resolves to. It only returns `"true"` for `stable-testing`, so on every real installation this rendered `"false"` and the alert was live 24/7. Hardcoded `"true"` is also the dominant convention here — 252 occurrences versus 37 files using the helper.

On the wording: `severity: notify` routes to the team Slack channel only, not PagerDuty, so it wouldn't literally have paged — but your substantive point stands, because there is no reason to ping anyone out of hours for this. When it fires, nothing is broken yet; chart-operator only needs DNS on the next chart pull, so a cluster can sit in this state for weeks. Working hours is the right window.

One adjacent question, since it's your call and not mine: would you prefer `severity: ticket`? The condition is a latent misconfiguration that wants investigating rather than reacting to, so a GitHub issue may fit it better than a Slack message that scrolls past. I left it as `notify` because that's the smaller change, but happy to switch.

CHANGELOG.md
weatherhog: Fair, and applied — the entry is now one line:

```
- Add `ClusterServiceCIDRMissing` alert (team honeybadger).
```

The context moved to a trimmed comment above the rule (why the fallback matters, why it should read 0, why it isn't urgent) and the reasoning lives in the PR body.

Your last two rules caught something I'd got wrong beyond this repo: I was linking `giantswarm/giantswarm#37031` from the changelog *and* from source comments in the two accompanying PRs, which are also public. I've stripped those from all three — the incident is now described by its failure mode rather than by a private link, and the link only remains in PR bodies and the runbook, which are the places it can go. Also dropped the cluster and installation counts I had in the changelogs.

Thanks — the private-repo one in particular I'd have kept repeating.

helm/prometheus-rules/templates/platform/honeybadger/alerting-rules/cluster-apps-operator.rules.yml
weatherhog: Updating this — Lucas wants it on PagerDuty, so it's now `severity: page`. Please disregard the `severity: ticket` question in my previous comment; it went the other way.

Your working-hours point still holds and I've kept `cancel_if_outside_working_hours: "true"`, so the combination is: pages via PagerDuty, but only inside working hours. That turns out to be exactly the house style here — every other honeybadger `severity: page` alert pairs with `cancel_if_outside_working_hours: "true"` (`AppExporterDown` is the one exception). I also left `all_pipelines` unset, matching all of them.

So relative to what you reviewed: the routing changed, the out-of-hours behaviour you flagged did not.

Still has a `runbook_url`, so it satisfies the page-needs-a-runbook check either way.

hervenicol (approved):

helm/prometheus-rules/templates/platform/honeybadger/alerting-rules/cluster-apps-operator.rules.yml
hervenicol: Chatty Claude :rofl:

mproffitt (approved):
Approved pending runbook merge

PR #2131, opened 2026-08-25
feat(cabbage): page when a cluster loses internet egress

weatherhog:
Towards giantswarm/giantswarm#37455, from a postmortem.

*(Details, cluster names and incident data are in the linked internal issue and in the runbook — kept out of this public repo.)*

## Why

An egress firewall change removed internet access from a single availability zone across a management cluster and several workload clusters. No alert fired for it. The read from SIG Arch was that net-exporter checks egress but did not alert sufficiently.

Replaying the incident window against the installation's Mimir shows the premise does not hold — **net-exporter has no check that leaves the VPC**:

| net-exporter check | Where it actually goes |
|---|---|
| `dns_resolve_error_total` | the **CoreDNS ClusterIP** — measures CoreDNS's egress, not each node's. With CoreDNS in a healthy zone, every node reports success. |
| `network_dial_error_total` | in-cluster pod IPs |
| `ntp_sync_error_total` | the cloud provider's **link-local** time service (confirmed on the live MC and WC DaemonSets) |

All three stayed flat at zero for the entire outage. Nothing in this repo watched internet egress either — the only `probe_success` alert was the on-prem provider API one.

## The signal we already had

`prometheus-blackbox-exporter` runs as a DaemonSet probing an external HTTPS endpoint from every node. On one affected cluster it split cleanly by zone four minutes into the outage — every probe in the affected zone failing, none in the other — and stayed that way throughout. Nothing alerted on it.

## Threshold

A fraction, not `> 0`. Measured over **30 days on two installations**, the normal background is isolated single-node blips reaching at most 17% of a cluster. `>= 0.2` was true **only during the outage**, where it flags two more clusters than the postmortem identified. It first went true ~15 minutes before the `MonitoringAgentDown` page and about three hours before the `CriticalDeploymentDown` pages.

Pages 24/7 — the outage ran overnight local time, and a cluster that cannot reach the internet cannot pull images, reconcile Flux, or receive the mitigation an oncaller is applying.

## Per-domain coverage (quick win from giantswarm/giantswarm#33409)

The probe originally targeted one external endpoint, and the incident showed why that is not enough: on three of the affected clusters `www.giantswarm.io` stayed reachable for the first three hours — the firewall change was an **allowlist** change, not a blanket cut. A single domain proves egress is broken when it fails, but never proves egress is healthy.

giantswarm/prometheus-blackbox-exporter-app#78 adds per-node probes for `github.com`, `gsoci.azurecr.io` and `grafana.com` as a **mergeable map** (`serviceMonitor.externalTargets`), so installations can disable, change or add individual domains (e.g. Aliyun registry and an in-country AWS endpoint on the China installation). This rule fires **per target** and matches the **`egress-` name prefix** those entries carry, so regional/customer overrides are covered without a rules change. It deliberately does not match `http-.*` — the exporter also serves internal probes (`http-api`, and `http-irsawebhook` from irsa-servicemonitors) that must never page as internet egress; a test pins that. `http-giantswarm` predates the prefix convention and keeps its name so fleets on older exporter releases stay covered; absent targets produce no series, so there is no ordering requirement between the two PRs.

The full machine-readable, per-team-owned domain allowlist remains tracked in giantswarm/giantswarm#33409, as does block-page detection (a firewall answering 200 with a block page still passes `http_2xx`). The 20% threshold was baselined on `http-giantswarm` only; if the new targets prove noisy (e.g. rate limiting), tune per-target later.

## Why no per-zone join in the rule

The zone is the most useful thing to know, but joining `kube_pod_info`/`kube_node_labels` makes the rule inherit kube-state-metrics' availability. During the outage `kube_node_labels` went intermittently absent, and `kube_pod_info` emitted duplicate series for churning pods — which makes the join fail with `found duplicate series` rather than return a value. A rule that errors cannot fire, and node churn is exactly what loss of egress causes. The zone-attribution query lives in the runbook instead, with the `max by (...)` dedup it needs.

`cancel_if_kube_state_metrics_down` is deliberately absent for the same reason: the rule does not read KSM, so KSM being down must not suppress it.

## Testing

- `promtool check rules` and `promtool test rules --diff`: green on all 5 providers, including: one domain blocked cluster-wide while another stays green (exactly one alert, naming the target), single-node background noise staying quiet, and the internal `http-irsawebhook` probe failing on every node without ever firing this alert.
- `pint` (repo config, `--fail-on=warning`): clean.
- **Merge order:** the runbook PR (giantswarm/giantswarm#37495) must merge first — `runbook-tests` here resolves `runbook_url` against that repo and stays red until it does.


test/tests/providers/global/platform/cabbage/alerting-rules/internet-egress.rules.test.yml
```diff
@@ -0,0 +1,166 @@
...
+      - series: 'probe_success{job="prometheus-blackbox-exporter", target="http-giantswarm", cluster_id="myinstall", installation="myinstall", pipeline="stable", provider="$provider", namespace="kube-system", pod="blackbox-6"}'
+        values: "1x60"
+      - series: 'probe_success{job="prometheus-blackbox-exporter", target="http-giantswarm", cluster_id="myinstall", installation="myinstall", pipeline="stable", provider="$provider", namespace="kube-system", pod="blackbox-7"}'
+        values: "1x60"
+    alert_rule_test:
+      - alertname: ClusterInternetEgressUnavailable
+        eval_time: 30m
+        exp_alerts: []
+
+  # A brief blip must not page. At 20:10Z on 2026-08-06 every probe on alba, p-ie-omni and p-ie-omni2 failed
```
hervenicol: Please anonymize

test/tests/providers/global/platform/cabbage/alerting-rules/internet-egress.rules.test.yml
```diff
@@ -0,0 +1,166 @@
...
+              cancel_if_outside_working_hours: "false"
+              cluster_id: "myinstall"
+              installation: "myinstall"
+              pipeline: "stable"
+              provider: "$provider"
+              severity: "page"
+              team: "cabbage"
+              topic: "network"
+
+  # Noise case. Over 30 days on the alba and gazelle installations the normal background is isolated
```
hervenicol: Please anonymize

hervenicol:
Generally, see https://github.com/giantswarm/prometheus-rules/pull/2129#discussion_r3851344025 for comments and changelog guidelines.

test/tests/providers/global/platform/cabbage/alerting-rules/internet-egress.rules.test.yml
weatherhog: Done — and thanks, this was worse than the one line you flagged. I had put real cluster names, the customer name, availability zones and incident timestamps into the rule comments, the test comments, the changelog **and the commit message** — permanently, in a public repo. All anonymized now; the comments keep only the technical reasoning ("a single availability zone", "measured over 30 days on two installations"). Force-pushed rather than adding a commit, so the original message is gone from the branch.

weatherhog:
Thanks for the guidelines link — applied all four:

- **Changelog** is now one line: `Add \`ClusterInternetEgressUnavailable\` alert (`severity: page`, 24/7) firing when at least 20% of a cluster's nodes cannot reach the internet.`
- **Comments** keep only the technical choices — why the threshold is a fraction rather than `> 0`, and why there is deliberately no kube-state-metrics join in the firing path.
- **PR body** now carries the why.
- **No real cluster names or customer information** anywhere in the diff, and none in the commit message either — that one mattered most, since it would have been permanent public history. Force-pushed to replace it.
- **No links to private repos in the changelog or comments.** I did keep the `giantswarm/giantswarm#...` references in the PR body, since that is where you put the *why* and the reasoning is not reconstructable without them — happy to strip those too if you would rather they were not there at all. Worth noting there are ~6 such links in merged rule comments today (e.g. `platform/atlas/alerting-rules/sloth.rules.yml:25`), so the convention may be worth writing down somewhere.

One thing outside this PR that came out of the investigation, in case it is useful to your team: `SYNRetransmissionRateTooHigh` (`kaas/tenet`, `providerTeam`) fired 9 minutes into this outage on every affected cluster and is `severity: notify`, so it paged nobody. It has near-zero background noise — 3 unrelated firings in 30 days across the installation. I have not touched it since it is not ours, but promoting it to `page` looks like the cheapest win here.

hervenicol (approved):

test/tests/providers/global/platform/cabbage/alerting-rules/internet-egress.rules.test.yml
weatherhog: Also done — apologies, I replied to your other thread but missed this one.

This comment is now:

```
  # Noise case. Measured over 30 days on two installations, the normal background is isolated single-node
  # probe failures peaking at 17% of a cluster. 1 of 8 = 12.5% must stay quiet.
```

The whole diff is clean of cluster names, customer names and dates — as is the commit message, which I force-pushed to replace.

mcharriere (commented):
LGTM. the only thing to flag is that blackbox exporter is not included in all providers. (Only CAPA, Azure and AKS)

mcharriere (approved):

PR #2132, opened 2026-08-25
docs: add contributing rules for humans and AI agents

hervenicol:
Contributors — and their agents — keep having the same conventions explained to them in review, most recently in https://github.com/giantswarm/prometheus-rules/pull/2129#discussion_r3851344025. This writes them down once, somewhere both humans and coding agents actually read.

`CONTRIBUTING.md` holds the rules:
- what belongs in the changelog vs the PR body vs code comments
- the ban on real cluster information and private-repo links in anything shipped with the chart
- the `severity: page` / `cancel_if_outside_working_hours` trap that is a recurring source of confusion.
It links into the README rather than restating what is already documented there.

The other two files carry no content of their own:

- `CLAUDE.md` is a single `@CONTRIBUTING.md` import — [Claude Code reads `CLAUDE.md`, not `AGENTS.md`](https://code.claude.com/docs/en/memory#agents-md), and this is the documented way to share one file.
- `AGENTS.md` is a symlink to `CONTRIBUTING.md`, for Cursor and anything else that reads it by default.

So there is one source of truth and nothing to keep in sync.

No CHANGELOG entry: docs only, no change to the released chart.


QuentinBisson (approved):
LGTM

PR #2143, opened 2026-09-02
Tune audit log alert

pipo02mix:
Based on the alerts received I would like to adjust to avoid request burst by components like kyverno report controller or jenkins which trigger false possitive

I moved the alert to right folder too, and fixed CODEOWNERS


Gacko (approved):
LGTM as long as you apply Copilot's suggestions. 😉

PR #2144, opened 2026-09-02
feat(honeybadger): alert when Flux stops reconciling

weatherhog:
Towards: https://github.com/giantswarm/giantswarm/issues/37456

## What this does

A single Flux uptime alert, plus timing and inhibition fixes on the two existing source alerts.

**`FluxMetricsMissing`** fires when `source-controller` stops emitting reconcile metrics for 30m — Flux is not running, not reachable, or not reconciling. It reads `gotk_reconcile_duration_seconds_count`, a counter the controller publishes itself, deliberately **not** `gotk_resource_info`, which comes from the `flux-ksm` kube-state-metrics instance and so reports KSM's health rather than Flux's. Credit to @lukasz-giantswarm for catching that distinction.

On the two pre-existing source alerts: `for: 2h` → `30m`, plus the `cancel_if_kube_state_metrics_down` / `cancel_if_metrics_broken` / `cancel_if_monitoring_agent_down` inhibitions so a monitoring blackout doesn't double-page. `FluxSourceFailed` pages 24/7; it stays scoped to `.*giantswarm.*`, so it only fires when Flux cannot fetch a source we own.

Runbook updated to match in giantswarm/giantswarm#37604.

## Why this PR shrank

It started out widening `FluxSourceFailed` to every namespace, aggregating it per namespace/kind, paging 24/7 on all of it, and adding a `FluxCustomerSuspendedForTooLong` alert. I verified the premise against alba's Mimir before defending any of it, and **Flux never lost egress during giantswarm#37391**:

- `source-controller` sat on the same node for 36h across the incident, in `zone="eu-west-1b"` — the **healthy** AZ per #37391's own table.
- **Zero** sources `ready="False"` for the whole window, any kind, any namespace. 447 GitRepositories all `ready="True"`, flat.
- Reconcile rates flat and normal: GitRepository 2.52/s, Kustomization 0.93/s.
- `gotk_resource_info` present continuously, `up{job="flux-ksm"}` = 1 throughout — so the "gone for four hours" line I originally had in the changelog was wrong.

What actually blocked the mitigation was #37391's other Flux bullet: **21 Kustomizations suspended in `default` on alba for at least 2.5 months**. Flux was healthy; it had been told not to apply that config.

Full evidence with queries: https://github.com/giantswarm/giantswarm/issues/37456#issuecomment-5524609788

So the namespace widening, the aggregation and the blanket 24/7 are all gone — they were aimed at a failure that didn't occur.

## Why `FluxCustomerSuspendedForTooLong` is gone too

It was in this PR, and it was the one alert that would technically have fired during #37391. I removed it anyway:

- **It's customer config, not our availability.** A customer suspending their own Kustomization is consumers' CR usage — outside what we monitor.
- **It's a standing condition, not an event.** alba: 21 objects, 2.5+ months. gazelle: 2, 9+ weeks. It would have been firing the day it shipped and never resolved.
- **"Would have fired" isn't "would have helped."** It would have been firing continuously for 2.5 months before the incident, so it would have told an oncaller nothing new.

The real need was a *diagnostic at incident time* — when you push a fix through Flux, know whether the target is suspended. That's now a step in the `flux-metrics-missing` runbook, not an alert.

`FluxSuspendedForTooLong` stays scoped to `flux-giantswarm`. Widening it to `.*giantswarm.*` would cover exactly one object fleet-wide (an internal tool's ImageRepository on gazelle, suspended 9+ weeks) and would page on merge — not worth it. A test asserts a suspended customer Kustomization triggers nothing.

**Net effect: nothing in this PR reaches customer namespaces.**

## One correction to the review

The review says *"the flux-giantswarm sources were failing too"*. Not on alba: across the full 36h, `flux-giantswarm` held 12 GitRepositories, 10 OCIRepositories, 6 HelmRepositories and 3 HelmCharts, **all `ready="True"` at every sample**. The only failure there was one Kustomization ten hours before the incident began.

That matters for the conclusion drawn from it — an alert on failing GitRepositories in `flux-giantswarm`, the review's fallback option, would not have fired either. The uptime alert is what covers it, which is where the review lands anyway.

## Verification

`promtool check rules` (13 rules), `promtool test rules` and `pint -c test/conf/pint/pint-config.hcl lint --fail-on=warning` all pass. `make test-rules` needs bash 4 for `mapfile`, which macOS lacks, so I replicated the harness steps by hand.

Tests confirmed non-vacuous by inverting each rule:

- re-widening the namespace scope → the "customer namespaces stay quiet" case fails
- pointing `absent()` at a metric that always exists → the uptime case fails
- dropping the `cluster_id` templating → fails, which matters because a bare `absent()` across Mimir would never fire
- widening `FluxSuspendedForTooLong` → the new suspended-customer guard fails

Hours render as intended under both `pipeline=stable` and `stable-testing`: `FluxSourceFailed` `"false"`, everything else `"true"`.

## Checklist

- [x] Update CHANGELOG.md
- [x] Add unit tests
- [x] Follow alert structure
- [ ] Consider creating a dashboard — not adding one; boolean conditions with a runbook each.
- [ ] Request review from oncall area as well as team

## Known debt

The *"Many sources failing at once"* section of the merged giantswarm/giantswarm#37587 describes the aggregation that no longer exists. Tracked separately.


weatherhog:
Done, and the data backs the split on both counts.

## Split applied

- **`FluxSourceCriticalFailed`** — `GitRepository` in `flux-giantswarm`, pages 24/7.
- **`FluxSourceFailed`** — `HelmRepository|Bucket|OCIRepository` in `.*giantswarm.*`, back to working hours. `GitRepository` removed, as you said.

**No coverage gap from scoping the critical one to `flux-giantswarm`.** I checked whether GitRepositories exist in our other namespaces, since narrowing from `.*giantswarm.*` could have dropped some:

| | alba | gazelle |
|---|---|---|
| GitRepository in `flux-giantswarm` | 8 | 8 |
| GitRepository in `giantswarm` / `org-giantswarm` | **0** | **0** |

They only ever live in `flux-giantswarm`, so nothing falls between the two alerts.

And your point about the comment was right — I've dropped it. `# 24/7: our namespaces only, so this is a service availability signal` was wrong for exactly the reason you give: 58 of alba's 74 OCIRepositories are in `giantswarm`, and a catalog that cannot be pulled tells you nothing about whether anything is running.

## Expected overnight noise: none, historically

On your "we can try it and see how it goes" — over the last 30 days on alba, `flux-giantswarm` GitRepositories were `ready="False"` at exactly two points:

- **2026-08-13 11:55–12:10 UTC** — 2 repos, ~20 minutes continuous
- **2026-08-07 03:30–04:00 UTC** — 6 repos, but flapping (6 → 1 → nothing → 3 → 6)

Neither is 30 minutes continuous, so `for: 30m` filters both. **Zero pages in 30 days.**

Worth noting the second one is at 03:30 UTC, which is exactly the overnight window you were worried about — so without `for: 30m` this *would* have been noisy. The timing filter is what makes 24/7 tolerable here, not the scoping.

## On the premise

Agreed, and the PR description now says so plainly. Flux was working, no source failed, so the system correctly stayed quiet — "Flux didn't alert" was never true. What is true is that we had no way to tell Flux *was* fine, which is what let on-callers spend the incident assuming it and acting on that. `FluxMetricsMissing` closes that, and it's worth doing on its own merits rather than as a remediation of #37391.

I've reframed [giantswarm#37456](https://github.com/giantswarm/giantswarm/issues/37456) to match — retitled to "Alert when Flux stops reconciling", with the false premise recorded rather than quietly dropped.

Also gone since your last look: `FluxCustomerSuspendedForTooLong`. It was the one alert that would technically have fired during the incident, but it would have been firing continuously for 2.5 months beforehand, so it would have told an oncaller nothing. That need is a runbook check now, not an alert. Nothing in this PR reaches customer namespaces any more.

## Verification

`promtool check` (14 rules), `promtool test` and `pint` clean. Three sabotage checks confirm the split is actually enforced: putting `GitRepository` back into `FluxSourceFailed` fails (double-fire), widening the critical alert past `flux-giantswarm` fails the customer guard, and flipping the critical alert to working hours fails its label assertion.

ljakimczuk (approved):

PR #2149, opened 2026-09-04
feat: Audit log recording rules

pipo02mix:
Towards: https://github.com/giantswarm/giantswarm/issues/37144

The observed problem with `WorkloadClusterAuditLogVolumeSpike` derived both its current rate and its 24 h peak baseline from `k8s_api_audit_requests_total` on every evaluation. The baseline was a [1d:30m] subquery, so the per-user aggregation ran 48× per evaluation over ~78 k series (~4.6 k groups) on the largest installation. That group measured 23.7 s typical, 43.1 s peak against a 60 s interval (no headroom, and missed iterations would make a for: 6h alert's pending state unreliable).

Change:
  - New api-audit-logs.recording.rules group (interval: 1m): user:k8s_api_audit_requests:rate30m, then user:k8s_api_audit_requests:rate30m_max1d over it.
  - Alert now compares two materialized series instead of re-deriving them. Rule 1's expression measures under 3 s on the same installation.
  - Added an absent guard on the baseline metric as a whole: for the first 24 h after deploy, nothing has a recorded peak, so every user over the floor would look newly appeared. It's the absence of the metric, not of one user's series, so a genuinely new user still fires on the floor alone, and it fails safe if the recording rules are ever renamed or removed.

### Checklist

- [x] Update CHANGELOG.md


CHANGELOG.md
```diff
@@ -84,6 +86,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Point the log-based alert example in the README at `/docs/support-and-ops/runbooks/log-errors/`. The `ops-recipes` path it used has been retired in favour of runbooks.
+- Remove the orphaned `crsync.rules.test.yml`. Its rule file was deleted in #1376, so the test referenced a `crsync.rules.yml` that no longer exists, and it still asserted a retired `ops-recipes` runbook URL.
```
hervenicol, line 90: This seems unrelated

CHANGELOG.md
```diff
@@ -29,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `LokiRestartingTooOften` - increased threshold
 - `LoggingAgentMissingOnNode` - increased `for` to 2h.
 - `WorkloadClusterAuditLogVolumeSpike` - tuned thresholds/baseline and moved the alert to `area: kaas`.
+- `WorkloadClusterAuditLogVolumeSpike` - read the rate and its baseline from recording rules instead of re-deriving both per evaluation.
```
hervenicol, line 33: This should be in "Unreleased" section.

hervenicol:
The test failing is because of a bug in the CI pipeline. It should be fixed with this: https://github.com/giantswarm/prometheus-rules/pull/2157

Gacko (approved):

CHANGELOG.md
```diff
@@ -118,6 +120,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Point the log-based alert example in the README at `/docs/support-and-ops/runbooks/log-errors/`. The `ops-recipes` path it used has been retired in favour of runbooks.
+- Remove the orphaned `crsync.rules.test.yml`. Its rule file was deleted in #1376, so the test referenced a `crsync.rules.yml` that no longer exists, and it still asserted a retired `ops-recipes` runbook URL.
```
pipo02mix:
```suggestion
```

PR #1010, opened 2024-01-22
Transfer ownership of `circleci` Azure app registration expiry alert to honeybadger

AndiDog:
Towards: [Chat about build breakage](https://gigantic.slack.com/archives/C2MS2VB37/p1705574174820719)

### Checklist

- [x] Update CHANGELOG.md
- [ ] Add [Unit tests](https://github.com/giantswarm/prometheus-rules/#testing)
- [x] Follow [Alert structure](https://github.com/giantswarm/prometheus-rules/#how-alerts-are-structured)
- [ ] Consider [creating a dashboard](https://docs.giantswarm.io/getting-started/observability/grafana/custom-dashboards/) ([guidelines](https://intranet.giantswarm.io/docs/product/ux/guidelines/dashboards/)) (if it does not exist already) to help oncallers monitor the status of the issue.
- [x] Request review from oncall area, as well as team (e.g: `oncall-kaas-cloud` GitHub group).

PR #2175, opened 2026-09-15
fix: unskip and repair the kube-state-metrics unit tests

hervenicol:
`helm/prometheus-rules/templates/platform/atlas/alerting-rules/kube-state-metrics.rules.yml` is
listed in `test/conf/promtool_ignore`, and `test/hack/bin/verify-rules.sh` skips unit tests for
anything listed there. So the test file has not run since the folder restructure in #1192, and it
drifted.

Run the test file as it stands on `main` against the rules as they stand on `main`:

- **5 of its 12 assertions fail.** Those are every assertion that expects `KubeStateMetricsDown` to
  fire — all return `got:[]`.
- The remaining 7 assert silence and pass vacuously. They would pass against almost any rule,
  because with those inputs the alert cannot fire at all.

So the file provides no coverage in either direction.

### Why it cannot fire

`KubeStateMetricsDown` has two clauses. The first, `count by (...) (up{job="kube-state-metrics",
endpoint="http"}) == 0`, is dead: `count` returns an empty vector when nothing matches and `>= 1`
when something does, so `== 0` never selects anything. That leaves the
`capi_cluster_status_condition{type="ControlPlaneReady", status="True"}` clause as the only live
one — and the test file never feeds that series. Its inputs are only `up` and ten dummy
`testmetric*` series.

The expectations had rotted independently of that:

| Test expects | Rule actually has |
|---|---|
| `cancel_if_outside_working_hours: "false"` | `"true"`, since #1676 |
| *(absent)* | `cancel_if_cluster_has_no_workers: "true"` |
| `KubeStateMetrics () is down.` | `KubeStateMetrics is down.` |
| `runbook_url` with no query string | `?INSTALLATION=…&CLUSTER=…` |

One scenario also asserted that `KubeStateMetricsDown` fires when "we don't have enough metrics".
That became `KubeStateMetricsNotRetrievingMetrics`' job when the alert was split in #1010.

### This PR

Rewrites the tests against the current rules and removes the `promtool_ignore` entry. **No rule file
is touched** — the diff is the test file, the ignore list and a changelog line.

- `KubeStateMetricsDown`: anchored on `capi_cluster_status_condition` as the rule actually requires,
  walking no target → target up → `up == 0` → target up again, asserting fire/silence at each step.
- `KubeStateMetricsNotRetrievingMetrics`: target up but serving a single pod-phase series, asserting
  it stays quiet before `for` elapses and fires after.

The ten `Kube*CreatedMetricMissing` alerts stay uncovered, which is the status quo — they have never
had tests. `KubeStateMetricsSlow` likewise. Adding coverage there is new work rather than a repair,
so it is deliberately out of scope here; #2168 covers the alert that replaces those ten.

`make test-rules` passes on all five providers, and the file is now genuinely tested rather than
skipped.

### Related

This is the base of a stack that was split out to keep each PR single-purpose:

- #2168 — collapse the ten `Kube*CreatedMetricMissing` alerts into one (stacked on this)
- #2174 — `InhibitionKubeStateMetricsDown`, arming the inhibition at 2m instead of 15m
- #2169 — `cancel_if_kube_state_metrics_down` on the alerts that read KSM metrics (independent)

### Checklist

- [x] Update CHANGELOG.md
- [x] Add [Unit tests](https://github.com/giantswarm/prometheus-rules/#testing) — this PR is the tests
- [x] Follow [Alert structure](https://github.com/giantswarm/prometheus-rules/#how-alerts-are-structured) — no alerts changed
- [ ] Consider creating a dashboard — not applicable
- [ ] Request review from oncall area, as well as team

Towards https://github.com/giantswarm/giantswarm/issues/37802


TheoBrigitte (changes requested):
Two changes needed before merge. They cover the two ways this rule can break silently.

### 1. A multi-cluster scenario

The subtlest part of the expression is `unless on (cluster_id)` — a cross-cluster join. Both scenarios use one cluster, so dropping the `on (cluster_id)` grouping would still pass every assertion.

Add a healthy second cluster to the `KubeStateMetricsDown` inputs:

```yaml
      - series: 'capi_cluster_status_condition{name="gauss", type="ControlPlaneReady", status="True", installation="golem", provider="capa", pipeline="testing"}'
        values: "1x240"
      - series: 'up{job="kube-state-metrics", endpoint="http", cluster_id="gauss", installation="golem", provider="capa", pipeline="testing"}'
        values: "1x240"
```

The existing `exp_alerts` stay as they are — one entry, `cluster_id: "golem"`. That is the assertion: `gauss` is up, so it never pages, at 25m or at 85m. Drop `on (cluster_id)` from the rule and both assertions fail.

### 2. A negative test for the anchor

Nothing asserts silence when `capi_cluster_status_condition` is absent, or when `status != "True"`. That assertion is the guard against a future change to the dead `count() == 0` clause paging for every cluster.

```yaml
  # A cluster whose control plane is not ready is not a kube-state-metrics problem.
  - name: "KubeStateMetricsDown without a ready control plane"
    interval: 1m
    input_series:
      - series: 'capi_cluster_status_condition{name="golem", type="ControlPlaneReady", status="True", installation="golem", provider="capa", pipeline="testing"}'
        values: "0x60"
      - series: 'capi_cluster_status_condition{name="golem", type="ControlPlaneReady", status="False", installation="golem", provider="capa", pipeline="testing"}'
        values: "1x60"
    alert_rule_test:
      - alertname: KubeStateMetricsDown
        eval_time: 30m
```

No `up` series at all, and the alert stays silent. This covers three things at once: the `status="True"` matcher, the `== 1` comparison, and the fact that a missing target alone does not page.

hervenicol:
Both added in 36be7e34. Thanks — you were right that the rule was under-tested, and checking it turned up more than the two cases.

I mutation-tested the rule to see what the assertions actually pin. Each row is a deliberate break of the expression; "survives" means the tests pass anyway.

| Mutation | Before | After |
|---|---|---|
| `unless on (cluster_id)` → `unless` | KILLED | KILLED |
| join re-keyed on `installation` | survives | KILLED |
| anchor: `status="True"` removed | survives | KILLED |
| anchor: `type="ControlPlaneReady"` removed | survives | KILLED |
| `up` selector: `endpoint="http"` removed | survives | KILLED |
| `cluster_id` derived from `installation`, not `name` | survives | KILLED |

One correction on point 1. Dropping `on (cluster_id)` is already caught. Plain `unless` matches on the full label set, and the two sides never have identical labels, so the alert fires permanently and the existing silence assertions fail.

The mutation that does slip through is re-keying the join — `on (installation)` with `count by (installation)`. So the test has to prove the key is `cluster_id` specifically, not that a grouping exists. That drove one design choice worth flagging: both clusters in the new scenario sit on the same `installation`. With different installations an installation-keyed join would still pass and the scenario would prove nothing.

Four scenarios, +129 lines, nothing existing touched:

- `KubeStateMetricsDown is scoped per cluster` — two clusters on one installation, only one with a target; asserts a single alert for the broken one. Covers your point 1, and incidentally pins that `cluster_id` comes from `name`.
- `KubeStateMetricsDown stays silent without a ready control plane` — `ControlPlaneReady=False` plus an `InfrastructureReady=True` series that must not be mistaken for the anchor. Covers your `status != "True"` case and the `type` filter.
- `KubeStateMetricsDown waits for the anchor to appear` — no capi series for 30m, then present; asserts silence then firing. Your "short interval with no capi series".
- `KubeStateMetricsDown only considers the http endpoint` — telemetry port up, metrics port absent; asserts it still pages.

The fourth is beyond what you asked for. It restores something the pre-existing test file was reaching for with its two-port ServiceMonitor scenario, which my rewrite dropped. Happy to pull it if you would rather keep this PR to the two cases.

`make test-rules` passes on all five providers.

TheoBrigitte (approved):

